PR #59, opened 2014-04-10
Fix WebServiceInjectionIntoNonContextualComponentTest

bafco:
add wsdl file to deployed archive


tremes:
I enabled the tests and removed "loadOnStartup = 1" parameter which I think is not neccessary here. JAX-WS related tests were revisited within CDITCK-457 so nothing else to merge here. Thank's anyway.

PR #94, opened 2015-07-22
Fix MultipleExceptionsInObserversNotificationTest - LondonRadioStatio…

jharting:
…n should be throwing the exception


tremes:
Merged. Thank you.

PR #95, opened 2015-07-23
BootstrapSEContainerTest - use try-with-resources for CDI

jharting:
We'll need to use try-with-resources or try/finally almost everywhere. Otherwise,
when a test fails the CDI container is not shut down and interferes with other test methods.


tremes:
Expanded and merged. Thank you.

PR #96, opened 2015-10-23
Include compatiiblity rules section as defined in JavaTM Technology C…

tremes:
…ompatibility Kit User’s Guide Template


tremes:
Updated & merged.

PR #97, opened 2015-11-13
Expand test for Conversation when invalidating session. Removed TODOs…

manovotn:
… regarding this

I modified test for conversation once the session is invalidated. This should precisely test when is the conversation destroyed. I also added another similar test with multiple conversation to verify similar behaviour for non-associated long-running conversations. Both tests should be relevant to CDI spec, chapter 6.7.4, TCK assertion qith id 'qa'
@tremes can you please review this?

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/context/conversation/servlet/ServletConversationTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/context/conversation/servlet/ServletConversationTest.java
@@ -19,8 +19,10 @@ package org.jboss.cdi.tck.tests.context.conversation.servlet;
 import static org.jboss.cdi.tck.TestGroups.INTEGRATION;
 import static org.jboss.cdi.tck.cdi.Sections.CONVERSATION_CONTEXT_EE;
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertTrue;
 
+import java.io.IOException;
 import java.net.URL;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


tremes:
Yes. It looks good to me. I can see two minor improvements wrt code itself :
1) Strings ("message: Hello",  "conversationCreated", etc. ) which occurs more than once could be stored in constants.  
2) I am rather going to comment inline.:)

impl/src/main/java/org/jboss/cdi/tck/tests/context/conversation/servlet/ServletConversationTest.java
```diff
@@ -184,6 +199,44 @@ public void testInvalidatingSessionDestroysConversation() throws Exception {
...
+    public void testInvalidatingSessionDestroysAllLongRunningConversations() throws Exception {
+        WebClient client = new WebClient();
+
+        // reset ActionSequence and begin two conversations
+        client.getPage(getPath("/resetSequence", null));
+        TextPage firstConversation = client.getPage(getPath("/begin", null));
+        String content = firstConversation.getContent();
+        String firstCid = getCid(content);
+        assertTrue(content.contains("message: Hello") && content.contains("transient: false"));
+
```
tremes: This 4 lines (maybe 5) above seem to appear quite often in this test. What about to extract them to some method? I think it would make the test little bit legible .

manovotn:
Updated PR:
* introduced constants for repeated String values
* introduced helper method for conversation initiation

tremes:
Merged. Thank you.

PR #101, opened 2016-03-29
CDITCK-512 tests for BeanConfigurator

manovotn:


impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/beanConfigurator/MonsterController.java
```diff
@@ -0,0 +1,76 @@
...
+    private static Ghost ghostInstance = new Ghost(false);
+    
+    // skeleton Supplier user as producer
+    public static Supplier<Skeleton> skeletonSupplier = () -> {
+        skeletonProducerCalled = true;
+        return new Skeleton(100);
+    };
+
+    // zombie Fuction used as producer
+    public static Function<Instance<Object>, Zombie> zombieProducingFunction = (Instance<Object> t) -> {
```
tremes: Casting `(Instance<Object> t)` is unnecessary I guess.

pom.xml
```diff
@@ -83,7 +83,7 @@
 
     <properties>
         <!-- CDI API -->
-        <cdi.api.version>2.0.Alpha1</cdi.api.version>
+        <cdi.api.version>2.0.Alpha2</cdi.api.version>
```
tremes: This will need a rebase.

impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/beanConfigurator/MonsterController.java
manovotn: This isn't a cast, it is an "explicit parameter type" declaration.
E.g. it should be the same as using just `(t)`

tremes:
Merged. Thank's

PR #104, opened 2016-04-07
Correct tests for BeanConfigurator.

manovotn:
Rewrote a lot of the previous content to avoid using *Builders* or to simplify one thing or another.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/bean/AfterBeanDiscoveryObserver.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/bean/AfterBeanDiscoveryObserver.java
@@ -16,6 +16,7 @@
  */
 package org.jboss.cdi.tck.tests.extensions.configurators.bean;
 
+import javax.enterprise.context.RequestScoped;
 import javax.enterprise.context.spi.CreationalContext;
 import javax.enterprise.event.Observes;
 import javax.enterprise.inject.spi.AfterBeanDiscovery;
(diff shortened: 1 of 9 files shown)


manovotn:
Just realized I cannot put 'work in progress' label or anything here so at least I can add a comment.

**Work in progress**

Currently it fails here (yet to be inspected) -> [BeanConfigurator#testInjectionPoints] (https://github.com/manovotn/cdi-tck/blob/fixBeanConfiguratorTests/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/bean/BeanConfiguratorTest.java#L106)

manovotn:
Updated the PR. In its current form, all test methods pass.
I changed the InjectionPoint test to check the IPs using ```Bean<T>.getInjectionPoints()```.

impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/bean/AfterBeanDiscoveryObserver.java
```diff
@@ -94,84 +91,49 @@ public void observeUndead(@Observes AfterBeanDiscovery abd, BeanManager bm) {
...
-            .type(Ghost.class);
-
-        InjectionPoint towerIP = injectionPointBuilder.build();
-
-        // add multiple IPs
-        ghost.addInjectionPoints(dungeonIP, towerIP);
+        ghost.scope(RequestScoped.class);
+
+        // test replacement of IPs
+        InjectionPoint ghostWeaponIP = null;
```
tremes: This variable is not necessary here (nothing will change when I remove it). I would make this a bit more readable and split this large `observeUndead` method into some separated methods (e.g configureSkeleton).

impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/bean/BeanConfiguratorTest.java
```diff
@@ -60,38 +65,51 @@ public void testCreationalAndDisposalMethods() {
...
-        
+
+        // instantiate Ghost and Vampire to verify their creational methods were called
+        Bean<Ghost> ghostBean = getUniqueBean(Ghost.class, Undead.UndeadLiteral.INSTANCE);
+        CreationalContext<Ghost> ghostCreationalContext = getCurrentManager().createCreationalContext(ghostBean);
+        Ghost ghost = ghostBean.create(ghostCreationalContext);
+
+        Bean<Vampire> vampireBean = getUniqueBean(Vampire.class, Undead.UndeadLiteral.INSTANCE);
+        CreationalContext<Vampire> vampireCreationalContext = getCurrentManager().createCreationalContext(vampireBean);
+        Vampire vampire = vampireBean.create(vampireCreationalContext);
```
tremes: These 3 lines could be reused in some private helper method instead of repeating them for each type.

tremes:
Otherwise it looks good!

impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/bean/BeanConfiguratorTest.java
manovotn: For Ghost and Vampire - yes, agree. For Skeleton and Zombie - not really. The reason is that I need a reference to their respective `CreationalContext` and their `Bean<T>` in order to call `destroy` methods.

manovotn:
Thanks for review.
I reflected the changes in the my latest update.

tremes:
Merged. Thank you!

PR #105, opened 2016-04-27
Stabilize MessageDrivenBeanInterceptorInvocationTest.

manovotn:
Came across a seemingly random failure of this test on some architecture. I suppose that might have been caused by race condition on a simple boolean value. Hence I turned it into LinkedBlockingQueue.
Similar was done with Weld test [here] (https://github.com/weld/core/commit/d429eaa88581ea1eb5c01100bcd7493e4edc9214).


tremes:
Merged. Thank you.

PR #109, opened 2016-08-02
JDK 9 has stricter inferring rules, adjust code accordingly.

manovotn:
Using JDK 9 EA (build 129), I came across a compilation error caused by generics usage.
JDK 9 probably has stricter rules for this although as of now, I haven't found any relevant info on that.


tremes:
Merged. Thank you.

PR #110, opened 2016-08-25
CDITCK-525 adding tests for when configurator and set methods are bot…

manovotn:
…h used within one observer method.


tremes:
This is failing at WildFly container (You can't check e.g ` assertTrue(POMExtension.ISE_CAUGHT);` in client test). Please update the test.

manovotn:
@tremes refactored, now runs on WFLY and belong to `integration` test group (runnable only `incontainer`). Uses servlet to gather results. Thanks for help, any other thoughts/required modifications?

tremes:
Merged. Thank you.

PR #111, opened 2016-10-17
Mark some configurators assertion as not testable.

tremes:
@mkouba @manovotn Guys if you have some idea how to test these assertions I am all ears.
It seems to me that it's not possible to get the original instance (IP, OM, BA) from configurator once you call `configure`.
Note that this is possible with `AnnotatedTypeConfigurator` where `getAnnotated` method is available.

### DIFF
--- a/impl/src/main/resources/tck-audit-cdi.xml
+++ b/impl/src/main/resources/tck-audit-cdi.xml
@@ -7419,7 +7419,7 @@
             <text>|setInjectionPoint()| replaces the |InjectionPoint|.</text>
         </assertion>
 
-        <assertion id="bba">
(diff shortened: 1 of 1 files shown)


manovotn:
Yea its not possible to retrieve the original, yet I don't think it would be useful to have that.
The only way to verify it would be to use identical IP in two beans and configure one of them (without triggering changes to it - just let the configurator overwrite it). Then make assertions that they "the same" - probably checking on qualifiers, type etc. Then again, I don't really know how to retrieve IP information once you leave the extension. So you might end up only checking the resulting injected beans...

tremes:
This will be tested. Thanks.

PR #116, opened 2016-11-21
CDITCK-535 update tests to use org.jboss.cdi.tck.util.ForwardingAnnot…

manovotn:
…ated util class.


impl/src/main/java/org/jboss/cdi/tck/tests/extensions/annotated/TestAnnotatedType.java
```diff
@@ -16,42 +16,54 @@
...
     public Set<AnnotatedField<? super X>> getFields() {
         getFieldsUsed = true;
-        return delegate.getFields();
+        return super.getFields();
     }
 
+    @Override
     public Class<X> getJavaClass() {
-        return delegate.getJavaClass();
+        return super.getJavaClass();
```
tremes: Couldn't we remove this method as well since it only delegates to predecessor? Otherwise looks good! Thanks

PR #124, opened 2016-12-08
CDITCK-539 test custom bean with interceptor.

tremes:


tremes:
Added @Priority to interceptor + fix testng assertion and merged.

PR #128, opened 2016-12-14
CDITCK-549 add test for extension registering custom OM which does no…

manovotn:
…t override notify method.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/lifecycle/broken/observerMethod/CustomObserverMethodWithoutNotifyMethodTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/lifecycle/broken/observerMethod/CustomObserverMethodWithoutNotifyMethodTest.java
@@ -1,0 +1,52 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/extensions/lifecycle/broken/observerMethod/ObserverMethodWithoutNotifyMethodTest.java
```diff
@@ -37,16 +39,37 @@
...
+    public static WebArchive createSecondTestArchive() {
+        return new WebArchiveBuilder().withTestClass(ObserverMethodWithoutNotifyMethodTest.class)
+                .withClasses(ExtensionAddingCustomObserver.class, Foo.class, FooObserver.class).withExtension(ExtensionAddingCustomObserver.class)
+                .withName(CUSTOM_OM + ".war").build();
+    }
+
+    @Test(dataProvider = ARQUILLIAN_DATA_PROVIDER)
     @SpecAssertions({ @SpecAssertion(section = AFTER_BEAN_DISCOVERY, id = "ec") })
     public void observerMethodNotOverridingNotifyMethodTreatedAsDefinitionError() {
+        deployer.deploy(CONFIGURATOR);
```
manovotn: @tremes not sure about this - do I need to call `undeploy` here? Since The deploy itself should not succeed I would guess not...

tremes: I can see compilation problem for `ExtensionAddingCustomObserver` but to your question - well maybe it would be safer to wrap this to try/catch block and do undeploy in the case of test failure (deployment will pass). This would also mean you can remove `@ShouldThrowException`.

manovotn: Uh true, forgot to add refactoring onto the commit, sorry.
Will fix this and change the test accordingly.

tremes:
Merged. Thank you!

PR #129, opened 2016-12-14
CDITCK-552 add ProducerConfigurator tests.

manovotn:
NOTE: this requires new CDI API + according changed to come classes to even compile

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/producer/BarProducer.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/producer/BarProducer.java
@@ -1,0 +1,60 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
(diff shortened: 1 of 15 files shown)


tremes:
Quite comprehensive tests but it looks ok at first glance. I will wait for the new API.

impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/producer/BarProducer.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Producer class. Contains loads of producer methods all of which are altered in extension.
+ *
+ * @author <a href="mailto:manovotn@redhat.com">Matej Novotny</a>
+ */
+@ApplicationScoped
+public class BarProducer {
+
+    public void ping() {
```
tremes: Just detail - this method is unnecessary right? Otherwise looks good.

manovotn: True, leftover from first version, will remove.

manovotn:
Rebased against master and took away unnecessary method.

tremes:
Merged. Thank you.

PR #131, opened 2016-12-15
CDITCK-559 Add test for InjectionTargetFactory#configure.

manovotn:
First shot at CDITCK-559, let's how that fares.


manovotn:
@tremes I haven't found the assertions there yet, so I created it along with the test. Please correct me if that was wrong.

tremes:
There is already one at https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/resources/tck-audit-cdi.xml#L6531 but you can definitely add the second one as you proposed.

manovotn:
@tremes you are right ofc, I had older TCK version and didn't see it there. I rebased it and split in into two separate assertions (as I had it originally in this PR). Is there anything else I should change/do?

tremes:
Merged. Thank you.

PR #136, opened 2017-01-13
Reenable CustomProviderTest now in SE group.

tremes:


impl/src/main/java/org/jboss/cdi/tck/tests/se/customCDIProvider/CustomCDIProviderTest.java
```diff
@@ -44,14 +47,11 @@
     }
 
     @Test
- //   @SpecAssertion(section = Sections.CDIPROVIDER_LOOKUP, id = "a")
+    @SpecAssertions({ @SpecAssertion(section = Sections.PROVIDER, id = "da"), @SpecAssertion(section = Sections.PROVIDER, id = "c") })
     public void testCustomCDIProvider() {
         CDI.setCDIProvider(new CustomCDIProvider());
```
tremes: Maybe we could define this as service provider directly in tested archive?

tremes: Hm it's true that it wouldn't likely reflect the assertion correctly.

mkouba: Yep, the purpose of `CDI.setCDIProvider()` is to avoid service providers.

mkouba:
@tremes Couldn't this test break other tests expecting the default CDIProvider impl?

tremes:
@mkouba Well that's why I put this test to SE group. I think (hope) the provider is set and used only during single container run which means only per one test in SE.

mkouba:
Ah, yes of course you're right!

tremes:
Merged.

PR #138, opened 2017-04-05
CDI-690 - update assertions and introduce new requestContext tests in

tremes:
SE.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/se/context/request/TestBean.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/se/context/request/TestBean.java
@@ -1,0 +1,52 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2017, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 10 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/se/context/request/TestBean.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.enterprise.inject.spi.BeanManager;
+import javax.inject.Inject;
+
+@ApplicationScoped
+public class TestBean {
+
+    @Inject
+    BeanManager bm;
+
+    private AtomicBoolean isReqContextActive = new AtomicBoolean(false);
```
mkouba: `isRequestContextActiveDuringPostConstruct` might be a better name...

tremes: Well yes. Thanks

impl/src/main/java/org/jboss/cdi/tck/tests/se/context/request/TestBean.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @PostConstruct
+    public void init(){
+        isReqContextActive.set(bm.getContext(RequestScoped.class).isActive());
```
mkouba: I would rather catch the `ContextNotActiveException` here and set the value to false. The reason is that the exception might lead to `CreationException` and that might complicate test debugging.

tremes: OK good point.

tremes:
Merged. Thanks for review.

PR #141, opened 2017-06-15
CDITCK-590 Add tests which assert order of event during bean discover…

manovotn:
…y for producer and managed bean.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/beanDiscovery/event/ordering/LifecycleEventOrderingTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/beanDiscovery/event/ordering/LifecycleEventOrderingTest.java
@@ -1,0 +1,67 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2017, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.cdi.tck.tests.extensions.beanDiscovery.event.ordering;
+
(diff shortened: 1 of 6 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/extensions/beanDiscovery/event/ordering/LifecycleEventOrderingTest.java
```diff
@@ -0,0 +1,89 @@
...
+        return new WebArchiveBuilder()
+            .withTestClassPackage(LifecycleEventOrderingTest.class)
+            .withExtension(ProductManagement.class).build();
+    }
+
+    @Inject
+    ProductManagement extension;
+
+    @Test
+    @SpecAssertions({
```
tremes: Minor - `@SpecAssertion` is now repeatable. So this is not necessary.

impl/src/main/java/org/jboss/cdi/tck/tests/extensions/beanDiscovery/event/ordering/ProductManagement.java
```diff
@@ -0,0 +1,85 @@
...
+ * @author <a href="mailto:manovotn@redhat.com">Matej Novotny</a>
+ */
+public class ProductManagement implements Extension {
+
+    private List<Object> listOfProducerEvents = new ArrayList<>();
+    private List<Object> listOfBeanEvents = new ArrayList<>();
+
+    // producer method observers
+    public void observePIP(@Observes ProcessInjectionPoint<PoorWorker, String> pip) {
+        listOfProducerEvents.add(pip);
```
tremes: Maybe we could take use of [ActionSequence](https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/util/ActionSequence.java) and store only class information to avoid this `instanceof` check.

manovotn: Not sure I follow. The actual class of `pip` will be some implementation's specific subclass implementing this interface (`ProcessInjectionPoint` in this case). Hence you basically need to check that whatever it is, it implements the given interface - so I went for `instanceof`.

tremes: You are right. So maybe we can go with something like https://github.com/tremes/cdi-tck/commit/0f52dde5237329778bd5ee703bbdb87cee83ea89. I consider it little bit more straightforward. Unfortunately this proposal fails because ActionSequence always checks the default sequence in this case so we would likely need to introduce new method to this ActionSequence class.

manovotn:
@tremes be it your way then, how do you like it now? :)

tremes:
Yes I like it:-) Thanks.

PR #161, opened 2018-03-26
CDITCK-612 Expand test to include removal based on Class<?>.

manovotn:


manovotn:
Expanded the test suite. It should now cover `contains()`, `containsAll()`, `removeAll()`, `retainAll()`.

PR #164, opened 2018-04-12
CDITCK-615 Regenerate signature tests for CDI API 2.0

hibell:
Generate sigtest file using sigtest-maven-plugin.


pom.xml
```diff
@@ -299,6 +299,16 @@
                         <arguments>-Drelease -Dtck-audit</arguments>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.netbeans.tools</groupId>
+                    <artifactId>sigtest-maven-plugin</artifactId>
+                    <version>1.0</version>
+                </plugin>
+                <plugin>
```
hibell: `sigtest-maven-plugin` can be omitted.

manovotn:
Thanks for the PR, I'll look into it next week.
I need to read up on this and see why we had those signatures for older Java versions attached (versus your approach of re-generating them on build).

hibell:
Hey @manovotn, I forgot to mention that I excluded these two classes from the signature file:
javax.enterprise.inject.se.SeContainer
javax.enterprise.inject.se.SeContainerInitializer

We think these classes should be challenged since they are for SE and not EE. Thanks.

hibell:
Hey @manovotn, I discussed with my team more and we will include the SE classes in our API jar. Thanks for the feedback. :)

manovotn:
I have never before dealt with signatures and sig testing but from what I can tell, you get slightly different signatures based on JDK you use, correct?

That would explain why previously there were 3 `.sig` files with jdk affixes.

Your solution makes it re-create on every build which means the result is dependent on what Java are you currently running and that might prove cumbersome.
Since the sig files are still the same for gives version of CDI, it seems more convenient to create them once and add them (like it was until now). Or am I missing something here?
I'll try to grab some more info about this...

manovotn:
Yet another problem is JDK 9/10 which the plugin doesn't seem to be able to handle. Not that we need signatures for that, but JDK 11 will arrive shortly and the problems will persist. The repo for that plugin was last updated in mid-2016.

BTW I tried to sketch something using your PR [here](https://github.com/manovotn/cdi-tck/tree/cditck615) - it's separated to another profile and also adds the generated file to the repository as I don't see a reason to re-generate it constantly. Though I only kept CDI API in that sig file. WDYT? Am I missing something?

hibell:
> I have never before dealt with signatures and sig testing but from what I can tell, you get slightly different signatures based on JDK you use, correct?

Yes, I believe you are right. It would also explain the JDK release being suffixed to the signature file as you mentioned.

> Another question - do you need to include JBoss versions of annotations and interceptors API in the signature? I did no see that in the previous .sig files.

For this, I tried to follow what was in the README that the TCK has for generating the signature files [here](https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/resources/readme_sigtest.txt):
```
java -jar  "/home/shane/java/sigtest-2.1/lib/sigtestdev.jar" Setup -classpath "/usr/local/java/jre/lib/rt.jar:/home/shane/.m2/repository/javax/enterprise/cdi-api/1.0/cdi-api-1.0.jar:/home/shane/.m2/repository/javax/inject/javax.inject/1/javax.inject-1.jar:/home/shane/.m2/repository/javax/el/el-api/2.1.2-b04/el-api-2.1.2-b04.jar:/home/shane/.m2/repository/org/jboss/interceptor/jboss-interceptor-api/3.1.0-CR1/jboss-interceptor-api-3.1.0-CR1.jar" -Package javax.decorator -Package javax.enterprise -FileName cdi-api.sig -static
```

> Yet another problem is JDK 9/10 which the plugin doesn't seem to be able to handle. Not that we need signatures for that, but JDK 11 will arrive shortly and the problems will persist. The repo for that plugin was last updated in mid-2016.

Hmm, that is interesting -- I did not think that would be an issue. I will try to investigate this further when I have time. Thanks for catching that. :)

> BTW I tried to sketch something using your PR here - it's separated to another profile and also adds the generated file to the repository as I don't see a reason to re-generate it constantly. Though I only kept CDI API in that sig file. WDYT? Am I missing something?

I like moving it to a profile. The BeanVal TCK generates the signature file using maven, so I thought adding it to the CDI TCK would be useful since it can be done automatically instead of having to remember to regenerate the signature files every new release.

Thanks for the feedback @manovotn. :]

manovotn:
>  I tried to follow what was in the README that the TCK
Uhm, I didn't know we have that file, will check. But I presume it needs to be added on the CP but maven plugin solves that for you automatically.

> instead of having to remember to regenerate the signature files every new release.

You only need to regenerate it for every **CDI** release. It's not like CDI gets released every month :-P
Hence I thought having the file baked directly into the repository makes sense.

If you want to go that way, feel free to cherry pick that commit on my branch over here and we can continue from there. Or if you have allowed edits from maintainers on this PR, I can just push it here. Let me know.

hibell:
> Uhm, I didn't know we have that file, will check. But I presume it needs to be added on the CP but maven plugin solves that for you automatically.

That makes sense.

> You only need to regenerate it for every CDI release. It's not like CDI gets released every month :-P
Hence I thought having the file baked directly into the repository makes sense.

That is true; my goal was just to have tooling in place to generate the file when it is necessary.

> If you want to go that way, feel free to cherry pick that commit on my branch over here and we can continue from there. Or if you have allowed edits from maintainers on this PR, I can just push it here. Let me know.

I think I did allow edits from maintainers so if you can push it here, that would be great. :) Thanks again.

manovotn:
@hibell Can you please verify that the `.sig`file (`impl/src/main/resources/cdi-api-jdk8.sig`) I attached to this PR works for your signature tests?
Of so, I can update the readme file to contain information about the added profile and can merge this.
Also, do you also need some sig files for JDK 9 or 10? They aren't really supported anywhere so I would wait for JDK 11 with that.

hibell:
@manovotn Sorry, I did not notice your request to test. I verified that the attached signature file `impl/src/main/resources/cdi-api-jdk8.sig` passes against Weld 3.0.3.Final:

```
$ mvn sigtest:check
[INFO] /Users/hwibell/cdi-tck-runner/target/surefire-reports/sigtest/TEST-cdi-tck-runner-1.0-SNAPSHOT.xml: 0 failures in /Users/hwibell/cdi-tck-runner/target/api-signature/cdi-api-jdk8.sig
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 6.195 s
[INFO] Finished at: 2018-05-08T02:02:50-04:00
[INFO] Final Memory: 7M/21M
[INFO] ------------------------------------------------------------------------
```

> Of so, I can update the readme file to contain information about the added profile and can merge 
    this.

That would be great. :) Here is an example [pom.xml](https://gist.github.com/hibell/a667eef7f86e536b9b7561f375122af7) that I used to run the CDI signature tests against Open Liberty if that is helpful.

> Also, do you need some sig files for JDK 9 or 10? They aren't really supported anywhere so I would 
   wait for JDK 11 with that.

We just needed the signature files for JDK 8 at the moment. Thanks for checking. :)

manovotn:
Going to merge this and then cherry-pick into 2.0 branch.
@hibell If there is something else you wanted to incorporate/change, please create another PR.

PR #168, opened 2018-04-23
CDITCK-618: Need to expose @Remove-annotated method to EJB interface

hibell:


hibell:
Thanks. The PR for 2.0 is done.

PR #175, opened 2018-05-18
CDITCK-622 Backport CDITCK-597, exclude EJBJarDeploymentTest

manovotn:


impl/src/main/resources/tck-tests.xml
```diff
@@ -52,6 +52,13 @@
                 </methods>
             </class>
 
+            <!-- CDITCK-622 -->
+            <class name="org.jboss.cdi.tck.tests.context.request.event.timeout.RequestScopeEventTimeoutTest">
```
mkouba: This should be `EJBJarDeploymentTest`, right?

manovotn: Yep, it is, I noticed it the second I sent the PR.
Good to know you're keeping an eye on it though ;-)

PR #182, opened 2019-01-09
Fix for issue 625 - backport for 2.0

doychin:
Use static atomic long to initialize id generator;

https://issues.jboss.org/browse/CDITCK-625
Signed-off-by: Doychin Bondzhev <doychin@dsoft-bg.com>

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/context/request/jaxrs/Foo.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/context/request/jaxrs/Foo.java
@@ -20,12 +20,15 @@ import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 import javax.enterprise.context.RequestScoped;
 import javax.inject.Inject;
+import java.util.concurrent.atomic.AtomicLong;
 
 @RequestScoped
 public class Foo {
 
     @Inject
     ObservingBean observingBean;
+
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/context/request/jaxrs/Foo.java
```diff
@@ -20,13 +20,16 @@
...
 import javax.inject.Inject;
+import java.util.concurrent.atomic.AtomicLong;
 
 @RequestScoped
 public class Foo {
 
     @Inject
     ObservingBean observingBean;
 
+    private static AtomicLong atomicLong = new AtomicLong(System.currentTimeMillis());
```
manovotn: Since you are basing the id off `getAndIncremenet()`, you no longer need to even use current time (or `Long` for that matter). You could simply initialize `AtomicInteger` with 0.

doychin: I know about that but I decided that the actual value does not mater and so a "pseudo-random" value for every run will also do the job.

manovotn: True, that was just a remark

manovotn:
Merged, thanks for contribution @doychin!

PR #183, opened 2019-03-26
CDITCK-626 Use Boolean as it implements same interfaces in JDK12

nziakova:
https://issues.jboss.org/browse/CDITCK-626
Also updating the text in tck-audit-cdi.xml for related section of the spec to match the text in the spec.


nziakova:
PR to 2.0 is here: https://github.com/cdi-spec/cdi-tck/pull/184

PR #184, opened 2019-03-26
CDITCK-626 Use Boolean as it implements same interfaces in JDK12

nziakova:
https://issues.jboss.org/browse/CDITCK-626
Also updating the text in tck-audit-cdi.xml for related section of the spec to match the text in the spec.

PR to master is here: https://github.com/cdi-spec/cdi-tck/pull/183


manovotn:
Thanks for the PR!

@nziakova can you please also update [`tck-tests.xml`](https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/resources/tck-tests.xml) and add this one test? It serves as an exclude list for tests which are expected to fail with the current version.

nziakova:
Done, thanks for the review!

PR #196, opened 2020-02-13
Update parent, update deps from javax to jakarta versions.

manovotn:
Some basic updates in regards to versioning plus the parent needed to grab the release profile.
FTR I think that most of the plugin versioning should eventually go to the parent, but for now I added it here to eliminate warnings during build.

I'm looking at the Jenkins jobs now, I'll probably create a new one that handles the release and does both, the dist build with upload and maven central release.

I think we could/should do a similar parent update followed by a release for EE 8 version too. Just to have it in place.

Related to https://github.com/eclipse-ee4j/cdi-tck/issues/195

### DIFF
--- a/impl/pom.xml
+++ b/impl/pom.xml
@@ -64,8 +64,8 @@
         </dependency>
 
         <dependency>
-            <groupId>javax.servlet.jsp</groupId>
-            <artifactId>jsp-api</artifactId>
+            <groupId>jakarta.servlet.jsp</groupId>
+            <artifactId>jakarta.servlet.jsp-api</artifactId>
         </dependency>
 
         <dependency>
@@ -95,12 +95,12 @@
 
         <dependency>
             <groupId>org.jboss.spec.javax.resource</groupId>
-            <artifactId>jboss-connector-api_1.6_spec</artifactId>
+            <artifactId>jboss-connector-api_1.7_spec</artifactId>
         </dependency>
 
         <dependency>
             <groupId>org.jboss.spec.javax.jms</groupId>
-            <artifactId>jboss-jms-api_1.1_spec</artifactId>
+            <artifactId>jboss-jms-api_2.0_spec</artifactId>
         </dependency>
 
         <dependency>
@@ -109,8 +109,8 @@
         </dependency>
 
         <dependency>
-            <groupId>javax.faces</groupId>
-            <artifactId>jsf-api</artifactId>
+            <groupId>jakarta.faces</groupId>
+            <artifactId>jakarta.faces-api</artifactId>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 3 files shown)


manovotn:
Oh apparently it has some build issues on JDK 11, I'll try that locally and see what we can do about that as we definitely want this to be viable on both, 8 and 11.

manovotn:
Ok, apparently JDK 9+ needs to have `com.sun.xml.ws:jaxws-ri:zip` added for some reason (I suppose some bits were extracted from the API part?). 
So I've added an auto-activated profile with that.

impl/pom.xml
```diff
@@ -95,12 +95,12 @@
...
         <dependency>
             <groupId>org.jboss.spec.javax.resource</groupId>
-            <artifactId>jboss-connector-api_1.6_spec</artifactId>
+            <artifactId>jboss-connector-api_1.7_spec</artifactId>
         </dependency>
 
         <dependency>
             <groupId>org.jboss.spec.javax.jms</groupId>
-            <artifactId>jboss-jms-api_1.1_spec</artifactId>
+            <artifactId>jboss-jms-api_2.0_spec</artifactId>
```
antoinesd: Shouldn't we get rid of these jobs spec in favor of jakarta ones?

pom.xml
```diff
@@ -99,21 +99,22 @@
...
-        <jboss.ejb.api.version>1.0.0.Final</jboss.ejb.api.version>
-        <jboss.jms.api.version>1.0.0.Final</jboss.jms.api.version>
-        <jboss.connector.api.version>1.0.1.Final</jboss.connector.api.version>
-        <jboss.annotations.api.version>1.0.1.Final</jboss.annotations.api.version>
-        <jboss.interceptors.api.version>1.0.1.Final</jboss.interceptors.api.version>
+        <jboss.ejb.api.version>2.0.0.Final</jboss.ejb.api.version>
+        <jboss.jms.api.version>2.0.0.Final</jboss.jms.api.version>
+        <jboss.connector.api.version>2.0.0.Final</jboss.connector.api.version>
+        <jboss.annotations.api.version>2.0.0.Final</jboss.annotations.api.version>
+        <jboss.interceptors.api.version>2.0.0.Final</jboss.interceptors.api.version>
```
antoinesd: Same question regarding jboss specs. What are their added value?

impl/pom.xml
manovotn: I don't know. I don't recall the original reason why we added these instead of javax. I think they are supersets of the original APIs or just 1:1. And they are in Central so I see no harm in them.

manovotn:
@antoinesd FYI I've sent similar bunch of changes to EE 8 branch here https://github.com/eclipse-ee4j/cdi-tck/pull/197

And I am also in process of setting up the release job on Jenkins - https://ci.eclipse.org/cdi/job/CDI%20TCK%20Release%20Job/

PR #218, opened 2020-12-04
Fix CustomCDIProviderTest by supplying an impl that doesn't return nu…

### LINKED ISSUE #217: CustomCDIProviderTest is incorrect
@starksm64 This change became incorrect after accepting the following change into cdi master - https://github.com/eclipse-ee4j/cdi/pull/445

The test asserts that you can return a provider which return `null` for its `getCDI()` call, but the code now guards against it.
I will send a fix for this shortly.

manovotn:
…ll on getCdi() calls.

Fixes #217 

This should get into next release before we hit Final.


Emily-Jiang (approved):
LGTM!

manovotn:
FTR, the JDK 11 Travis build fails because of the issue I reported here - https://github.com/eclipse-ee4j/cdi-tck/issues/219

PR #221, opened 2021-01-05
Update TestNG to 7.3; perform changes needed to execute tests with th…

### LINKED ISSUE #220: Update testng version and relevant bits
Going through some dep updates in Weld, I figured I cannot update TestNG from 6.x to 7.x because newer versions by default disable loading DTD from unsecure urls. And the files are declared in CDI TCK, namely [here](https://github.com/eclipse-ee4j/cdi-tck/blob/master/impl/src/main/resources/tck-tests.xml) and [here](https://github.com/eclipse-ee4j/cdi-tck/blob/master/impl/src/main/resources/tck-tests-previous.xml).

There are a few more tweaks needed in TCKs to update testng, such as custom listener extending a removed interface but I think master TCKs should be kept up to date instead of staying on an older version.

manovotn:
…is version.

Fixes #220 

Note that doing this means any implementation will need to use TestNG 7.x to execute the testsuite. However, this version comes with the TCK itself, so it shouldn't be an issue.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/impl/testng/ProgressLoggingTestListener.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/impl/testng/ProgressLoggingTestListener.java
@@ -22,7 +22,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.testng.IInvokedMethod;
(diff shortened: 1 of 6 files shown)


manovotn:
This has been up for over two months with no comments or reviews - I am going to merge it.

PR #224, opened 2021-03-16
Introduce BeansXML API for beans.xml file creation.

### LINKED ISSUE #222: Update beans.xml files to version 3.0
The CDI TCK currently uses the Shrinkwrap API to construct its test applications, including dynamic generation of metadata files like beans.xml.  This results in all of the beans.xml files in the TCK being stuck at version 1.1:

`<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" bean-discovery-mode="all" version="1.1" xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd"/>`

With the release of Jakarta EE 9 and the switch to the jakarta namespace, these should all be updated to version 3.0.

manovotn:
Change tests to use this API instead of shrinkwrap descriptors where applicable.
Correct existing beans.xml files to use new namespaces and version 3.0.
Add a simple test of BeansXml API.

Fixes #222

This PR:
* Removes usage of Shrinkwrap descriptors for `beans.xml`
  * Note that we still use that for other descriptors which should be amended at some later point and the dependency could than be dropped
* Changes (hopefully) all `beans.xml` to use the new namespace and version
* Adds new `BeansXml` API that can construct a `beans.xml` asset which can be added into the constructed test archive

FTR I ran this version against latest Weld (on WildFly) to verify that it works.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/bindings/aroundConstruct/ConstructorInterceptionTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/bindings/aroundConstruct/ConstructorInterceptionTest.java
@@ -23,14 +23,12 @@ import static org.jboss.cdi.tck.interceptors.InterceptorsSections.INVOCATIONCONT
 import static org.jboss.cdi.tck.util.ActionSequence.assertSequenceDataEquals;
 
 import jakarta.enterprise.inject.Instance;
-
 import org.jboss.arquillian.container.test.api.Deployment;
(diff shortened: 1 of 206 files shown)


manovotn:
Expanded existing `BeanDiscoveryTest` to also test an explicit bean archive with `beans.xml` in version `1.1` and `2.0`

PR #240, opened 2021-06-10
Update CDI API to 4.0.0.Alpha1; rework tests that use now removed APIs.

### LINKED ISSUE #239: Update CDI API to 4.0.0.Alpha1, rework tests using deprecated methods
CDI API 4.0.0.Alpha1 removed some long deprecated methods.

CDI TCK are still using those (`BM.fireEvent()`, `@New` tests etc) so we need to rework those tests and cut a release which can be consumed by implementors.

manovotn:
Fixes #239 

This updates the CDI API to latest version and then reworks lots of tests that were using now removed APIs such as `BM.fireEvent`.
All the tests that were making assumptions on the behavior of `@New` were removed as the annotation is no longer in CDI.

NOTE - I am in process of testing this with Weld. Once I am sure the set of changes is complete, I will switch this from draft to actual PR.


manovotn:
As part of this PR I have also updated Travis to use JDK 11 and 16 (instead of 8 and 11).

starksm64 (approved):

starksm64:
Is this ready to merge?

manovotn:
> Is this ready to merge?

I didn't get to test this with WFLY + Weld as WFLY needed a bunch of updates that I didn't manage to finish yet (other EE APIs use the now removed CDI API calls making it crash hard). I was going to get back to this tomorrow if that's ok?
Otherwise, I have tested this without EE container and in that case, all tests passed for me - so if you really need it right away, I am 80% confident this is a complete set of changes we need :-)

starksm64:
Ok, I'll just work off the current master.

manovotn:
Updated with few more findings from when I ran against WFLY + Weld. This should now be a complete set.

@starksm64 feel free to merge this when you need it.

impl/src/main/java/org/jboss/cdi/tck/tests/event/fires/FireEventTest.java
```diff
@@ -212,8 +211,8 @@ public void testInjectedEventCanHaveBindings() {
     }
 
     /**
-     * This test verifies that binding types can be specified dynamically when firing an event using {@link Event#fire()} by
-     * first using the {@link Event#select()} method to retrieve an Event object with associated binding types.
+     * This test verifies that binding types can be specified dynamically when firing an event using {@code Event#fire()} by
+     * first using the {@code Event#select()} method to retrieve an Event object with associated binding types.
```
Ladicek, line 215: Not sure why, but probably doesn't hurt.

impl/src/main/java/org/jboss/cdi/tck/tests/event/metadata/EventMetadataTest.java
```diff
@@ -67,14 +67,8 @@ public void testSimpleEvent(SimpleEventNotifier notifier, SimpleEventObserver ob
...
 
-        notifier.fireSimpleEventBeanManager();
-        verifyMetadata(observer.getSyncMetadata(), false, SimpleEvent.class, Any.class);
-
         notifier.fireSimpleEventWithQualifiers();
         verifyMetadata(observer.getSyncMetadata(), true, SimpleEvent.class, Alpha.class, Bravo.class, Any.class);
-
-        notifier.fireSimpleEventBeanManagerWithQualifiers();
-        verifyMetadata(observer.getSyncMetadata(), false, SimpleEvent.class, Charlie.class, Any.class);
     }
```
Ladicek, line 72: Why not test `EventMetadata` when firing the event through `BeanManager.getEvent()`? I get it, it wasn't tested before :-)

impl/src/main/java/org/jboss/cdi/tck/tests/event/fires/FireEventTest.java
manovotn: Just do that IDE doesn't complain because in order to use `@link`, you should also specify parameter types which, to me, make the text less readable. So I just went with `@code`, no real difference though :)

impl/src/main/java/org/jboss/cdi/tck/tests/event/metadata/EventMetadataTest.java
manovotn: Well, I didn't want to introduce new assertions, just adapt existing tests and remove those that test now removed behavior.
That being said, I am not against it. Although I am not sure you can expect anything else than a `non-null` value for injection point (which is what the test asserts on). So if you want to contribute that, feel free to add it on top of this PR or even in a separate one - whatever floats your boat :-)

Ladicek (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/event/fires/FireEventTest.java
Ladicek: Ah good point re parameter types, thanks!

impl/src/main/java/org/jboss/cdi/tck/tests/event/metadata/EventMetadataTest.java
Ladicek: I'll try that when this PR is merged :-)

PR #245, opened 2021-07-22
merge spi changes for cdi-lite into master

starksm64:


starksm64:
Recreating from a clean branch

PR #246, opened 2021-07-22
merge spi changes for cdi-lite into master 

starksm64:


starksm64:
The ECA check is broken, following details and validating my email verifies the ECA exists:
ValidECA

PR #251, opened 2021-08-12
Correct service entries

dstepanov:


starksm64:
I had to recreate this by merging it.

Issue #258, opened 2021-08-24
Add a new cdi-full test group


starksm64:
We need to distinguish between test common to both CDI-lite and the full CDI specification requirements. Discussions concluded that the best way to do this would be via a new cdi-full test group to mark tests that make use of features like @Interceptors, @Decorator, passiviation, portable extensions, etc.

PR #261, opened 2021-08-25
Add the CDI_FULL (cdi-full) test group, #258

starksm64:


impl/src/main/java/org/jboss/cdi/tck/TestGroups.java
```diff
@@ -30,6 +31,11 @@
  */
 public final class TestGroups {
 
+    /**
+     * Integration test - cannot run in embedded arquillian container
```
Ladicek, line 35:
```suggestion
     * Test for a CDI Full feature. Doesn't apply to CDI Lite.
```

Maybe?

PR #266, opened 2021-08-26
Replace package-private scope with public

dstepanov:


starksm64:
@dstepanov Try the https://github.com/eclipse-ee4j/cdi-tck/pull/277 PR to see if that works for Micronaut.

Issue #276, opened 2021-09-04
Need to support a two pass run mode to properly deal with tests run against annotation processor output


starksm64:
The current approach that attempts to reload classes generated from annotation processing is producing class loading issues, so we need to use a two pass approach that allows the test source to be passed to an annotation processor via the SourceProcessor SPI, and then run the tests against the resulting updated tests.

manovotn:
This issue is now outdated, closing it.

PR #277, opened 2021-09-04
Update how SourceProcessor is used to a two pass mode

starksm64:
Addresses #276


impl/pom.xml
```diff
@@ -427,18 +431,44 @@
...
+                                </property>
+                                <property>
+                                    <name>surefire.testng.verbose</name>
+                                    <value>10</value>
+                                </property>
+                            </properties>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
```
Ladicek, line 471: Just curious, this isn't really needed here, is it? This will have to be done by _consumers_ of the TCK, right?

starksm64: Yes, that is used by the impl/src/test tree in the org.jboss.cdi.tck.test.spi.* packages to have some validation of the two pass/annotation processor behavior.

Ladicek:
This looks sensible to me, as much as I can understand it, but I think @dstepanov is the best one to review.

dstepanov:
I'm going to look at it soon

dstepanov:
This looks like a big mishmash. 

Do I understand correctly that the tests are supposed to fail on the first run and also generate some classes which we include on the second run?

I think it would be better just to unzip the sources jar and include it as a test source directory...

dstepanov:
I have tried my solution but the annotation processor detects valid errors and breaks.

starksm64:
No, the tests should not even be run on the first pass because of this addition to the org.jboss.cdi.tck.AbstractTest:

```java
    /**
     * Override the Arquillian IHookable method to skip invocation of the test during a two pass
     * run mode needed by implementations that use a SourceProcessor and annotation processing phase.
     * @param callback - testng callback to run test
     * @param testResult - test result object to pass to callback
     */
    @Override
    public void run(IHookCallBack callback, ITestResult testResult) {
        SourceProcessor.Phase phase = SourceProcessorSuite.currentPhase();
        if(phase != null && phase == SourceProcessor.Phase.PASS_TWO) {
            super.run(callback, testResult);
        } else {
            // Skip test run in PASS_ONE
        }
    }
```

Only the static archive creation method annotated with @ Deployment should be run on the first pass.

starksm64:
> I think it would be better just to unzip the sources jar and include it as a test source directory...

This will not work in general because there is not a one to one mapping between a package and what is included in a test. An annotation processor will see more classes than are actually available in the test archive. Maybe tests need to be restructured to have a simple mapping between package and test archive.

dstepanov:
I will investigate more why does it run tests for me

impl/pom.xml
```diff
@@ -86,6 +86,11 @@
             <version>2.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
dstepanov, line 89: This should be deleted

dstepanov:
Looks like I made it work, at least for phase one.

The problem with the code above is that it doesn't work for the tests that are expecting the exception to be thrown:
```
org.testng.TestException: 
Method SimpleBeanSpecializationTest.testProducerMethodOnSpecializedBeanNotCalled()[pri:0, instance:org.jboss.cdi.tck.tests.inheritance.specialization.simple.SimpleBeanSpecializationTest@53841544] should have thrown an exception of type class jakarta.enterprise.inject.UnsatisfiedResolutionException
```

Ladicek:
I wasn't present on the initial discussion that lead to this `SourceProcessor` thing, but my opinion is that this approach is fragile and we should abandon it.

Everything related to compilation with annotation processing and then running the code can (AFAIK) and should (IMHO) be done entirely in an Arquillian adapter. The only thing the CDI TCK must do is provide the source code in the `Archive` instances returned by `@Deployment` methods, in addition to the compiled classes. And we do that now, I think?

The Arquillian adapter would then ignore the compiled classes in the `Archive`, take the sources, compile them as it wishes, and run the application. The application could be executed either in a separate classloader, or in an external JVM. Quarkus does the classloader thing, while all the adapters for Jakarta EE servers out there do the external process thing. (Thorntail also did the external process thing.) Both options entail an "interesting" set of troubles, and I understand it is a massive pain to implement an Arquillian adapter in general, but it's the only scalable solution we have.

starksm64:
Any integration at the Arquillian level simply does not work because it is based on the classes under test. Look for how to unit test an annotation processor and you will find frameworks that introduce access to the compiler api to test code fragments. Nothing does what we are trying to do.

I'll put this topic on the agenda for the next dev call. We probably should just drop the SourceProcessor notion and do as Denis suggested, just have annotation processor based impls unpack the source code, run their source processing phase, and then run the unit tests from that location.

Ladicek:
I'm like 99% sure that Arquillian's ability to take the deployment archive, perform arbitrary modifications to it, send it over network to an application server, and then orchestrate test runs there, is more than enough to cover this situation. Of course, assuming that the deployment archive includes source code. An archive processor can ignore the compiled classes, unpack the sources, compile them in an arbitrary way, and add the compiled classes back to the archive.

Of course, I didn't really try doing this, it's all based on my understanding of Arquillian, so I may easily be wrong.

starksm64:
The problem is not all adaptors are actually using serialization of the test archive. Also, this is just really fighting how annotation processing works with existing java compilers. I'm now in favor of just making this a TCK runner configuration detail and removing all knowledge of annotation processing from the TCK helper classes.

Ladicek:
I'd argue that all Arquillian adapters are essentially fighting how the target container is supposed to be used. Surprisingly, it often works. (When it doesn't, it's "fun", no doubt about that.)

Now, I'm no longer sure that the deployment archives the TCK tests produce include source code. If they don't, what I'm saying is of course nonsense.

starksm64:
@dstepanov see the latest update to this PR that allows exceptions generated during calls to the SourceProcessor to be seen during runs of the TCK in phase two. The only change you would need to make is this new code block as seen in the 
org.jboss.cdi.tck.test.spi.container.MockContainer to your org.jboss.arquillian.container.spi.client.container.DeployableContainer implementation:

```java
    @Override
    public ProtocolMetaData deploy(Archive<?> archive) throws DeploymentException {
        System.out.printf("MockContainer.deploy, %s\n", archive.getName());
        // Handle exceptions generated in the PHASE_ONE source processor run
        if(archive.contains(ArchiveBuilder.SP_EXCEPTION)) {
            Node exNode = archive.get(ArchiveBuilder.SP_EXCEPTION);
            ExceptionAsset exceptionAsset = (ExceptionAsset) exNode.getAsset();
            Exception ex = exceptionAsset.getSource();
            DeploymentException dex;
            if(ex instanceof DeploymentException) {
                dex = (DeploymentException) ex;
            } else {
                dex = new DeploymentException("SourceProcessor failure", ex);
            }
            throw dex;
        }
        // TODO, deployment classloader to validate generated classes?
        ShrinkWrapClassLoader classLoader = new ShrinkWrapClassLoader(archive.getClass().getClassLoader(), archive);
       ...
}

manovotn:
This PR should now be irrelevant since we chose a different approach - closing it.

Issue #262, opened 2021-08-25
Use CDI_FULL test group to exclude @Interceptors tests


starksm64:
Breaking up the sets of tests into separate categories, first @Interceptors based tests need to added to the CDI_FULL test group.

starksm64:
I think we are good to close this.

graemerocher:
@manovotn @starksm64 @Ladicek what was the final decision on `@AroundInvoke` in the target class. I see definitions like:

https://github.com/eclipse-ee4j/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/contract/aroundInvoke/Bean3.java

Are included with the `lite` tests?

manovotn:
@graemerocher you are right, this isn't supposed to be in Lite.
 `cdi-tck/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/` is where all interceptor specification TCK tests are. I think I reworked most interceptors tests to operate on interceptors with bindings and we didn't want to split interceptors TCK into different packages.

I must have missed this one (and possibly others?), it was lots and lots of tests...
Looking at the assertions, it seems to be testing this:
```
        <group>
            <text>Around-invoke methods can have |public|, |private|, |protected|, or |package| level access.</text>
            <assertion id="ca">
                <text>Test |public| level access.</text>
            </assertion>
            <assertion id="cb">
                <text>Test |private| level access.</text>
            </assertion>
            <assertion id="cc">
                <text>Test |protected| level access.</text>
            </assertion>
            <assertion id="cd">
                <text>Test |package| level access.</text>
            </assert
```

Which means this TCK test can be rewritten to use the standard interceptor with bindings.
But we should then keep a copy of this test (or verify there is other similar) for CDI Full.

graemerocher:
@manovotn thanks for the clarification. There seem to be several cases, like this one as well:

https://github.com/eclipse-ee4j/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/order/aroundInvoke/Vehicle.java

manovotn:
Yeah, most of the reworks I did were focused around `@Interceptors` and other stuff that was `grep`-able.

The one you mentioned will need to be marked Full and we'll need to make a similar version for Lite with only the supported bits (so that we don't coverage on this).

manovotn:
Feel free to send PRs or, at the very least, please create new issue(s) for them so we keep track of all of them.

PR #278, opened 2021-09-05
First pass at adding cdi-full to interceptors tests

starksm64:
Part 1 of #262

Moves @ Interceptors tests under org.jboss.cdi.tck.tests.full.interceptors.* package
Add cdi-full group to tests
Adds bean defining annotations needed for these tests to pass weld master

[INFO] Tests run: 45, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.319 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 45, Failures: 0, Errors: 0, Skipped: 0


impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/aroundConstruct/AbstractInterceptor.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jboss.cdi.tck.interceptors.tests.contract.aroundConstruct;
+package org.jboss.cdi.tck.tests.full.interceptors.contract.aroundConstruct;
```
Ladicek, line 17: Not a request for change, just a question. Some interceptors tests are currently present in `org.jboss.cdi.tck.interceptors.tests`, while others are in `org.jboss.cdi.tck.tests.interceptors`. I _think_ the first ones are for features defined by the Interceptors spec, while the second ones are for features added by CDI. I wonder if we want to keep that distinction?

impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/lifecycleCallback/Bar.java
```diff
@@ -14,7 +14,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jboss.cdi.tck.interceptors.tests.contract.interceptorLifeCycle;
+package org.jboss.cdi.tck.tests.full.interceptors.contract.lifecycleCallback;
```
Ladicek, line 17: This is once again Git not being able to figure out what's going on. The other side of this is marked `Wai3Hesi`.

impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/interceptorLifeCycle/Bar.java
```diff
@@ -14,7 +14,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jboss.cdi.tck.interceptors.tests.contract.lifecycleCallback;
+package org.jboss.cdi.tck.tests.full.interceptors.contract.interceptorLifeCycle;
```
Ladicek, line 17: This is once again Git not being able to figure out what's going on. The other side of this is marked `Wai3Hesi`.

impl/src/main/java/org/jboss/cdi/tck/tests/interceptors/definition/interceptorOrder/InterceptorOrderTest.java
```diff
@@ -59,20 +59,4 @@ public void testInterceptorsCalledInOrderDefinedByBeansXml(Foo foo) {
...
-
-        transaction.execute();
-
-        List<String> sequence = ActionSequence.getSequenceData();
-        assertEquals(sequence.size(), 4);
-        assertEquals(sequence.get(0), AnotherInterceptor.class.getName());
-        assertEquals(sequence.get(1), TransactionalInterceptor.class.getName());
-        assertEquals(sequence.get(2), Transaction.class.getName());
-        assertEquals(sequence.get(3), AccountTransaction.class.getName());
-    }
```
Ladicek, line 77: I think this test file doesn't need to be split and should be moved to `full` entirely. The 1st test (kept in this file in this PR) tests `beans.xml` enablement, the other test (moved to `full` in this PR) uses both `@Interceptors` and `@AroundInvoke` on bean class.

impl/src/main/java/org/jboss/cdi/tck/tests/interceptors/invocation/InterceptorInvocationTest.java
```diff
@@ -83,32 +80,6 @@ public void testInitializerMethodsNotIntercepted() {
...
-        DependentInstance<Wheat> bean = newDependentInstance(Wheat.class);
-        Wheat instance = bean.get();
-        assertNotNull(instance);
-        AlmightyInterceptor.methodIntercepted = false;
-        bean.destroy();
-
-        assertTrue(WheatProducer.destroyed);
-        assertTrue(AlmightyInterceptor.methodIntercepted);
-    }
-
```
Ladicek, line 111: Both these tests should IMHO stay out of `full`, they still apply to Lite I think?

Ladicek (approved):
Looks fine in general, though I'd appreciate a second pair of eyes on the comments about test splitting.

impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/aroundConstruct/AbstractInterceptor.java
starksm64: That is a possibility, and would we introduce another test group to distinguish these as not being in lite, or simply label them full?

impl/src/main/java/org/jboss/cdi/tck/tests/interceptors/invocation/InterceptorInvocationTest.java
starksm64: The WheatProducer in this test uses @Interceptors though, so it cannot be in lite.

impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/aroundConstruct/AbstractInterceptor.java
Ladicek: No I think I'd label these tests `CDI_FULL` as usual. I was only commenting on the package structure.

impl/src/main/java/org/jboss/cdi/tck/tests/interceptors/invocation/InterceptorInvocationTest.java
Ladicek: Ah I see, right. I wonder if we're losing coverage then. I think I'd rather change `WheatProducer` to use an interceptor binding, as producer/disposer interception is certainly part of Lite.

impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/aroundConstruct/AbstractInterceptor.java
manovotn: Yeam might be good to keep the package structure. It indeed looks like `org.jboss.cdi.tck.interceptors.tests` is basically an interceptors TCK.

impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/lifecycleCallback/Bar.java
manovotn: I guess that's what you get for naming half of your classes either `Foo` or `Bar` :D

manovotn (changes requested):
Most of these interceptor tests actually need to stay in Lite because they test functionality that is common to both. So instead of moving them to Full, we need to adapt them not to use `@Interceptors` and apply interceptor bindings instead.
I have added comments to relevant tests.

starksm64:
Ok, I'll redo the PR to keep org.jboss.cdi.tck.interceptors.tests and replace the use of @Interceptors in most locations. I guess the only @Interceptors test to keep for full would be those explicitly adding multiple interceptors?

manovotn:
> @interceptors test to keep for full would be those explicitly adding multiple interceptors?

I think I added comments to all the test classes that need changing. Some of those classes where you only marked certain methods as Full were IMO correct.

impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/aroundConstruct/AbstractInterceptor.java
manovotn: BTW I am seeing similar issue while looking at tests using portable extensions so I thought it might be sensible to place any interceptor TCK tests that require CDI full to `org.jboss.cdi.tck.interceptors.tests.full` package while keeping any CDI-related tests under the package that you used in this PR.
E.g. if a test needs to be moved, do it in following way:
* [interceptor TCK test] `org.jboss.cdi.tck.interceptors.tests` -> `org.jboss.cdi.tck.interceptors.tests.full`
* [CDI TCK test] `org.jboss.cdi.tck.tests.interceptors` -> `org.jboss.cdi.tck.tests.full.interceptors`

starksm64:
I'm going through the tests, and I'm not sure what is being tested in org.jboss.cdi.tck.tests.interceptors.definition.lifecycle.order.LifecycleInterceptorOrderTest.

It looks like it is testing that "If an interceptor is enabled for the application and for the bean archive, then the enablement from the bean archive is ignored by the container.", so that the ordering of the AnotherInterceptor and TransactionalInterceptor should be 
1. AnotherInterceptor because it has an @Interceptors binding which is higher than @InterceptorBinding
2. TransactionalInterceptor because it has an @InterceptorBinding

It seems like this test should be updated to use @InterceptorBinding and @Priority and kept in Lite. The fact that the beans.xml interceptors is used will be ignored in lite, and since it is also ignored in full, the test should work in both?

manovotn:
@starksm64 from what I can tell, the test tries to assert these two statements - https://github.com/eclipse-ee4j/cdi-tck/blob/master/web/src/main/resources/tck-audit-cdi.xml#L5060-L5072

None of the two is related to Lite as we:
* Do not support interceptors declared on target class
* Do not support usage of `@Interceptors`

> It looks like it is testing that "If an interceptor is enabled for the application and for the bean archive, then the enablement from the bean archive is ignored by the container."

I am not seeing that anywhere in `LifecycleInterceptorOrderTest`. One interceptor is using `@Interceptors` which doesn't require enablement and the other is via bindings and enabled in `beans.xml` file.

So unless I am missing something, I would say this belong to Full group.

starksm64:
Superseded by #285. All changes suggested on this issue should be in that PR.

Issue #265, opened 2021-08-25
Use CDI_FULL to exclude portable extension tests


starksm64:
Breaking up the sets of tests into separate categories, portable extension based tests need to added to the CDI_FULL test group.

manovotn:
Two parts are sent, reviewed and merged and I am hoping to post the remaining part today or tomorrow. 
Most of it should hopefully be just moving tests to `full` package as they already reside in packages suggesting tests purely for portable extensions (therefore with no test coverage loss for Lite)

manovotn:
I've sent the last part of changes. If accepted, this issue can IMO be resolved.

PR #284, opened 2021-09-15
Move/rework/alter tests using CDI extensions, part 2

manovotn:
Partly addressing #265

This is just Part 2 and still to be continued although the rest of tests seem to be inside `impl/src/main/java/org/jboss/cdi/tck/tests/extensions/` package meaning they should be relatively easy to just move under Full group.
See https://github.com/eclipse-ee4j/cdi-tck/pull/283 for part 1.


impl/src/main/java/org/jboss/cdi/tck/tests/full/deployment/discovery/VerifyingExtension.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jboss.cdi.tck.tests.deployment.exclude;
+package org.jboss.cdi.tck.tests.full.deployment.discovery;
```
Ladicek, line 17: This is a Git hiccup I guess?

impl/src/main/java/org/jboss/cdi/tck/tests/full/deployment/exclude/VerifyingExtension.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jboss.cdi.tck.tests.deployment.discovery;
+package org.jboss.cdi.tck.tests.full.deployment.exclude;
```
Ladicek, line 17: Ah yea, here's the counterpart.

Ladicek (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/full/deployment/discovery/VerifyingExtension.java
manovotn: Yep, I certainly didn't do any such wild renames :-)

PR #286, opened 2021-09-16
Move/rework/alter tests using CDI extensions, part 3

manovotn:
Another bunch of file changes related to #265 and still not done.

Apart from the `BeanByTypeTest`, the rest is literally just slapping on the test category and changing packages which is why the number of files changed grows so fast. So nothing much to review here but I still cut it else the number of files would be even scarier :)

Will continue tomorrow and add another commit here or create a new PR based off this one to keep track of progress in `org.jboss.cdi.tck.tests.extensions` package.


impl/src/main/java/org/jboss/cdi/tck/tests/extensions/bean/bytype/BeanByTypeTest.java
```diff
@@ -30,6 +30,7 @@
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.cdi.tck.AbstractTest;
 import org.jboss.cdi.tck.shrinkwrap.WebArchiveBuilder;
+import org.jboss.shrinkwrap.api.BeanDiscoveryMode;
```
Ladicek, line 33: Not needed?

Also, why is this even in the `extensions` package? It doesn't use extensions at all, does it? Is that just because `BeanManager` is in the `spi` package, or something like that?

Ladicek (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/extensions/bean/bytype/BeanByTypeTest.java
manovotn: Don't know. I was thinking the same but didn't really know where to put it at the moment so I just left it there.

manovotn:
I'll merge this.
I might change the package of `BeanByTypeTest` in one of the later parts along with other (if any) tests that would be under `extension` package but relate to behavior that Lite has as well.

PR #288, opened 2021-09-17
Move/rework/alter tests using CDI extensions, part 5

### LINKED ISSUE #265: Use CDI_FULL to exclude portable extension tests

manovotn:
Last part which, I hope, resolves #265 
With this, all tests under `extension` package should be addressed.

In this PR, some tests were moved but some were split as they contained Lite related tests as well.
I also changed package names for these Lite tests as the `extension` prefix made little sense.

Once again, each altered test was verified against Weld.


impl/src/main/java/org/jboss/cdi/tck/AbstractTest.java
```diff
@@ -46,6 +47,7 @@
  */
 public abstract class AbstractTest extends Arquillian {
 
+    // TODO this will have to be eventually changes as Lite implementations do not have BeanManager, only BeanContainer
```
manovotn, line 50: I am aware of this TODO and left it here deliberately.
We will have to alter this abstract class somehow so it works for Lite and Full implementations (as in Lite you cannot inject `BeanManager`).

Ladicek (approved):

impl/src/main/java/org/jboss/cdi/tck/AbstractTest.java
Ladicek: One option is we could tighten up the spec a bit. Currently, we say that even injecting (or otherwise obtaining) a `BeanManager` is non-portable -- we could perhaps remove that. We'd keep the provision that calling `BeanManager` methods (those that are not inherited from `BeanContainer`) is non-portable, obviously.

manovotn: Yeah, that would work fine. I didn't want to solve it as part of this commit though so I only left a TODO note.

Issue #290, opened 2021-09-20
More decorator related tests need the cdi-full test group


starksm64:
In looking at current core tests failures related to missing bean defining annotations, I came across the org.jboss.cdi.tck.tests.implementation.builtin.metadata.broken.injection.decorated package tests. These are using @Decorated and validating the requirements around that qualifier. 

Another test is org.jboss.cdi.tck.tests.implementation.builtin.metadata.broken.injection.BuiltinDecoratorInjectionTest

starksm64:
@manovotn can you look at the org.jboss.cdi.tck.tests.lookup.injectionpoint.{InjectableReferenceTest,InjectionPointTest} test. There are several decorators in there, but they don't seem to be tested. Not sure if those tests should move to cdi-full, so I left them alone.

manovotn:
Both decorators are in fact used. However, the one in `InjectableReferenceTest` can be replaced by interceptor (as it tests something unrelated to decorators).
The other test needs to be partly moved to Full.
I have sent a PR that addresses these two tests - https://github.com/eclipse-ee4j/cdi-tck/pull/292

PR #291, opened 2021-09-20
Move more decorator related tests to cdi-full, #290

starksm64:
This addresses the remaining decorator tests.


impl/src/main/java/org/jboss/cdi/tck/tests/full/decorators/implementation/builtin/metadata/BuiltinMetadataBeanTest.java
```diff
@@ -14,8 +14,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jboss.cdi.tck.tests.implementation.builtin.metadata;
+package org.jboss.cdi.tck.tests.full.decorators.implementation.builtin.metadata;
```
Ladicek, line 17: The test doesn't seem to be strictly about decorators. IMHO, the addition here should be just `full`, not `full.decorators`.

Ladicek: This comment actually applies to all files changed in this PR. A lot of the changed tests are not only about decorators, and those that specifically test decorators already have `decorator` in the package name, so with this change, the package name of those tests will contain the word `decorator` twice.

Ladicek:
Looks OK baring the one comment about package names.

web/src/main/java/org/jboss/cdi/tck/tests/implementation/builtin/metadata/session/YoghurtInterceptor.java
```diff
@@ -0,0 +1,42 @@
...
+import jakarta.interceptor.InvocationContext;
+import org.jboss.cdi.tck.tests.full.decorators.implementation.builtin.metadata.Frozen;
+
+@SuppressWarnings("serial")
+@jakarta.interceptor.Interceptor
+@Frozen
+public class YoghurtInterceptor implements Serializable {
+
+    @Inject
+    private Bean<org.jboss.cdi.tck.tests.full.decorators.implementation.builtin.metadata.YoghurtInterceptor> bean;
```
manovotn, line 19: Imports and FQCN of some classes in this interceptor are wrong. It should reference classes in this package instead.

manovotn (commented):
I agree with Ladislav's comments that the package change should be purely addition of `full` into it. I also added one other comment that I believe is an error.
Other than that, it looks good.

PR #292, opened 2021-09-21
Rewrite/adjust more tests using decorators

manovotn:
Related to #290 

This fixes the two tests @starksm64 asked me to take a look it. Feel free to put your work on top of this PR or just merge it and create a new one.

About `InjectionPointTest` - this test actually asserts things like `AnnotatedParameter` and others which we won't use in Lite. However, we want to use `InjectionPoint` API as such so I think we want to keep them in for now.


manovotn:
I didn't realize there is https://github.com/eclipse-ee4j/cdi-tck/pull/291 already up. I guess we can just merge both of these separately :)

Ladicek (approved):
Took me a little while to grasp, but yea, looks good!

Issue #289, opened 2021-09-20
Address missing bean defining annotations in Lite portion of tests


starksm64:
With the non-Lite tests moved under the full package and labelled with the cdi-full test group, the Weld jboss-tck-runner emedded tests show the following failure count due to missing bean defining annotations:

[ERROR] Tests run: 1423, Failures: 385, Errors: 0, Skipped: 245

The next step in #273 is to add the missing bean defining annotations to obtain a clean run of the core tests under Weld.

manovotn:
@starksm64 I was planning on doing this once your PR was merged. If you want, I can take it over.

starksm64:
@manovotn sure, go ahead. I am currently working on #290 as I saw several decorator related tests using the Decorated qualifier that were not moved to cdi-full. I'm half way through that, so a PR will be coming shortly so you can avoid updating those tests.

manovotn:
Alright, I will start working on this tomorrow :)

starksm64:
@manovotn with the latest changes it looks like we are down to 95 errors with Weld on core tests:
```
[ERROR] Tests run: 1274, Failures: 95, Errors: 0, Skipped: 79
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:09 min
[INFO] Finished at: 2021-09-23T06:42:51-05:00
[INFO] ------------------------------------------------------------------------
```

You mentioned being off at some point this week, so just assign this issue to me if you don't finish before you leave.

manovotn:
@starksm64 yea, I have covered most at this point and I plan on finishing that today or tomorrow :-)
BTW I will be off next Tue through Fri.

PR #296, opened 2021-09-22
Address missing bean defining annotations, part 2.

manovotn:
Part two of #289 

This part is purely addition of bean defining annotations.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/abortProcessing/ObserverExceptionAbortsProcessingTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/abortProcessing/ObserverExceptionAbortsProcessingTest.java
@@ -18,6 +18,7 @@ package org.jboss.cdi.tck.tests.event.observer.abortProcessing;
 
 import static org.jboss.cdi.tck.cdi.Sections.OBSERVER_NOTIFICATION;
 
+import jakarta.enterprise.context.Dependent;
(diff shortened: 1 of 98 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/abortProcessing/ObserverExceptionAbortsProcessingTest.java
```diff
@@ -39,6 +40,7 @@ public static WebArchive createTestArchive() {
     public static class AnEventType {
     }
 
+    @Dependent
```
Ladicek, line 43: We should move bean classes (as well as the event type) out of the test class. Not needed in this PR I guess.

manovotn: I have noticed this several times throughout TCKs. While I am with you in that it is nicer to have them separate, this is a perfectly legitimate way to declare beans. Maybe having a few of these in TCKs provides us with some on-the-side testing of this fact :-)

Ladicek: I actually think it's uglier to have the classes outside of the test class :-), but we need the classes outside so that testing an annotation processor based implementation is even possible. We debated that on the calls several times. But, again, not a problem of this PR!

PR #304, opened 2021-10-26
remove SourceProcessor and everything related

Ladicek:
The `cdiLiteMode` flag is retained, because that seems to be useful.
Everything else should be gone with this commit.


manovotn (approved):
Quick-scanned this PR and it looks fine. I trust you were thorough :)

Ladicek:
I tried :-)

PR #308, opened 2021-10-29
Update CDI API, introduce test for Instance.Handle.

### LINKED ISSUE #303: Provide test coverage for programmatic lookup improvements
Test coverage issue for https://github.com/eclipse-ee4j/cdi/issues/521

We should be able to use some tests that are already existing in [Weld](https://github.com/weld/core/tree/master/tests-arquillian/src/test/java/org/jboss/weld/tests/instance/enhanced).

manovotn:
Fixes #303 

Introduces a test for `Instance.Handle` that asserts bean metadata, resolution and bean destruction (copied from Weld TS).
Also updates CDI API to latest version.
Lastly, I removed a `TODO` note in `AbstractTest` because we can now happily keep using `BeanManager` in these tests.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/lookup/dynamic/handle/InstanceHandleTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/lookup/dynamic/handle/InstanceHandleTest.java
@@ -1,0 +1,139 @@
+package org.jboss.cdi.tck.tests.lookup.dynamic.handle;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+
+import jakarta.enterprise.context.Dependent;
+import jakarta.enterprise.inject.Instance;
+import jakarta.enterprise.inject.spi.Bean;
+import jakarta.enterprise.inject.spi.BeanManager;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.cdi.tck.AbstractTest;
+import org.jboss.cdi.tck.shrinkwrap.WebArchiveBuilder;
+import org.jboss.cdi.tck.util.ActionSequence;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.jboss.test.audit.annotations.SpecVersion;
+import org.testng.annotations.Test;
+
+import java.util.Iterator;
+import java.util.List;
+
+@SpecVersion(spec = "cdi", version = "4.0")
+public class InstanceHandleTest extends AbstractTest {
+
+    @Deployment
+    public static WebArchive createTestArchive() {
+        return new WebArchiveBuilder().withTestClassPackage(InstanceHandleTest.class).build();
+    }
+
+    //@SpecAssertion(section = PROGRAMMATIC_LOOKUP, id = "TODO")
+    @Test
+    public void testIsResolvable() {
+        Client client = getContextualReference(Client.class);
+        ActionSequence.reset();
+        assertNotNull(client);
+        assertTrue(client.getAlphaInstance().isResolvable());
+        assertFalse(client.getBigDecimalInstance().isResolvable());
+    }
+
+    //@SpecAssertion(section = PROGRAMMATIC_LOOKUP, id = "TODO")
+    @Test
+    public void testGetHandle() {
+        Client client = getContextualReference(Client.class);
+        BeanManager beanManager = getCurrentManager();
+
+        ActionSequence.reset();
+        assertNotNull(client);
+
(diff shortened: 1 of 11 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/lookup/dynamic/handle/InstanceHandleTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+        ActionSequence.reset();
+        assertTrue(instance.isAmbiguous());
+        for (Iterator<Instance.Handle<Processor>> iterator = instance.handles().iterator(); iterator.hasNext(); ) {
+            try (Instance.Handle<Processor> handler = iterator.next()) {
+                handler.get().ping();
+            }
+        }
+        assertEquals(4, ActionSequence.getSequenceSize());
+        ActionSequence.assertSequenceDataContainsAll("firstPing", "secondPing", "firstDestroy", "secondDestroy");
```
Ladicek, line 112: How does ordering work here? I see `Processor` has `getPriority`, but that isn't called anywhere?

manovotn: @Ladicek it is undefined.
Which is why the assertion only checks for presence, not for ordering.

The `getPriority()` indeed isn't tied to `@Priority` anyhow. I copied the test from Weld and this seems to be a redundant method on the bean. I have already removed it.

Ladicek: Ah gotcha! I'm not familiar with `ActionSequence` :-)

impl/src/main/java/org/jboss/cdi/tck/tests/lookup/dynamic/handle/InstanceHandleTest.java
```diff
@@ -0,0 +1,114 @@
...
+        assertEquals(1, sequence.size());
+        assertEquals(alpha2Id, sequence.get(0));
+
+        alpha1.destroy();
+        // Subsequent invocations are no-op
+        alpha1.destroy();
+
+        sequence = ActionSequence.getSequenceData();
+        assertEquals(2, sequence.size());
+        assertEquals(alpha1.get().getId(), sequence.get(1));
```
Ladicek, line 75: Here, you're calling `alpha1.get()` after `alpha1` has been `destroy()`ed, do we really guarantee that you will get the previous instance?

Ladicek:
OK, just one more question, otherwise LGTM.

impl/src/main/java/org/jboss/cdi/tck/tests/lookup/dynamic/handle/InstanceHandleTest.java
manovotn: This is a very good point and the behavior in the test is incorrect. Spec currently doesn't say what happens but there are basically two options:
* Throw exception if the `Handle` already resolves and subsequently destroyed the instance
* Create a new instance

The first seems like a way to go so long as we document it properly.
I will send the PRs this afternoon.

impl/src/main/java/org/jboss/cdi/tck/tests/lookup/dynamic/handle/InstanceHandleTest.java
```diff
@@ -0,0 +1,139 @@
...
+        alphaHandle.destroy();
+        // verify that the destruction happened
+        List<String> sequence = ActionSequence.getSequenceData();
+        assertEquals(1, sequence.size());
+
+        // try to invoke Handle.get() again; this should throw an exception
+        try {
+            alphaHandle.get();
+            fail("Invoking Handle.get() after destroying contextual instance should throw an exception.");
+        } catch (IllegalStateException e) {
```
Ladicek, line 109: I assume you're going to submit a separate PR to CDI for this?

Ladicek (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/lookup/dynamic/handle/InstanceHandleTest.java
manovotn: This is now fixed in this PR.
CDI API PR is here - https://github.com/eclipse-ee4j/cdi/pull/555

impl/src/main/java/org/jboss/cdi/tck/tests/lookup/dynamic/handle/InstanceHandleTest.java
manovotn: https://github.com/eclipse-ee4j/cdi/pull/555

manovotn:
Pushed a tiny update that removed my leftover authorship on one of the files and changed variable naming in test from `handler` to `handle`.

PR #310, opened 2021-11-02
Initial TCK for the language model

Ladicek:


Ladicek:
There's a couple of TODOs in the code, plus I want to add one or two other smaller items, but this PR should have like 90% of what's required, IMHO.

graemerocher:
So there is no actual test framework involved, the docs will say to run it from your test framework of choice?

Ladicek:
That's the idea, yes.

Ladicek:
Added a commit with some more tests (for a bunch of corner cases, I especially enjoyed figuring out how inherited repeatable annotations are supposed to work). I think the tests in this PR now have pretty good coverage.

There are still 8 TODOs in the code. The TODOs are of 3 kinds:

- 1 TODO for testing records -- but I have no idea how
- 2 TODOs for unbounded wildcard types -- they currently don't test what the spec says, but I tend to believe that the spec should change
- 5 TODOs where I'm not exactly sure what the correct behavior is

Ladicek:
Added one more commit that resolves some TODOs and adds some more tests.

Summary of the 5 remaining TODOs:

- 1 TODO for testing records -- but I have no idea how
- 2 TODOs for unbounded wildcard types -- they currently don't test what the spec says, but I tend to believe that the spec should change
- 2 TODOs for annotations on transitive type parameter bounds

manovotn (approved):
LGTM, great job!

graemerocher (approved):

Ladicek:
Added two more commits for some strange cases (in particular, annotations and type annotations on parameters of enum constructors, which are notorious for having synthetic parameters).

By now, there is only 1 TODO remaining: how to test support for records.

Ladicek:
Rebased on current `master` which uses CDI 4.0.0.Beta2 and hence the split `lang-model` module.

manovotn:
The README change looks good!
Going to merge this since it was already approved and we discussed during the meeting that it is ready.

PR #311, opened 2021-11-04
Introduce tests for @Priority declared on stereotypes.

### LINKED ISSUE #282: Provide test coverage for @Priority used on a stereotype
Provided https://github.com/eclipse-ee4j/cdi/issues/495 gets accepted, we will need test coverage for it.

Namely we need tests that:
* Use `@Stereotype` with `@Priority` to enable interceptor
* Use `@Stereotype` with `@Priority` to enable decorator (Full only!)
* Use `@Stereotype` with `@Priority` to enable alternative
* Have a `@Alternative` bean declaring two stereotypes both of which have `@Priority` and the bean itself doesn't declare `@Priority`
* Have a `@Alternative` bean declaring two stereotypes both of which have `@Priority` and the bean itself does declare `@Priority`
* Transitive stereotype scenario (stereotypes declaring stereotypes) where all stereotypes declare `@Priority` and assertion that the most specific value is taken
* ??

manovotn:
Fixes #282 

Tests are taken from Weld PR here - https://github.com/weld/core/pull/2559
I also added one extra test (`testSeveralPrioritiesInSingleStereotypeHierarchy`) that works with several declarations of `@Priority` in a hierarchy of single stereotype. We don't explicitly define this in spec text, but it is implied by annotation inheritance and overriding.


Ladicek (changes requested):
Otherwise looking good.

impl/src/main/java/org/jboss/cdi/tck/tests/definition/stereotype/priority/DeltaAlt2.java
```diff
@@ -0,0 +1,14 @@
+package org.jboss.cdi.tck.tests.definition.stereotype.priority;
+
+import jakarta.enterprise.context.Dependent;
+import jakarta.enterprise.inject.Alternative;
+
+@Dependent
+@Alternative
+@AnotherPriorityStereotype // this will give it @Priority(200)
```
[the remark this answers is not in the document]
manovotn: I know, I chose the latter and this test should reflect it.
And knowing that it can be contentious is the very reason I wanted to add this test to TCKs - to have at least some ground for it if it is reviewed at some later point.

manovotn:
Incorporated javadoc suggestions, the only remaining questionmark is https://github.com/eclipse-ee4j/cdi-tck/pull/311#discussion_r743484271

Ladicek:
Just to repeat it here: maybe add a test or two with an `@Inherited` stereotype? :-)

manovotn:
> Just to repeat it here: maybe add a test or two with an `@Inherited` stereotype? :-)

True, we could add at least a simple test to see if the working case (single `@Priority) is inherited properly.
Quickly scanning TCKs, I can see we test this for scopes declared on stereotypes, but that's about it. But maybe I am just not seeing it.

manovotn:
@Ladicek this PR now reflects the changes made to specification and tests for definition exception if a single stereotype indirectly declares more than one priority value. I think we are good to go now.

impl/src/main/java/org/jboss/cdi/tck/tests/definition/stereotype/broken/multiplePriorities/ConflictingPrioritiesFromSingleStereotypeTest.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * A single stereotype which indirectly declares multiple different priorities.
+ */
+@SpecVersion(spec = "cdi", version = "4.0")
+public class ConflictingPrioritiesFromSingleStereotypeTest extends AbstractTest {
+
+    @Deployment
+    @ShouldThrowException(DefinitionException.class)
+    public static WebArchive deploy() {
+        return new WebArchiveBuilder().withTestClass(ConflictingPriorityStereotypesTest.class)
```
Ladicek, line 23: I guess you meant

```suggestion
        return new WebArchiveBuilder().withTestClass(ConflictingPrioritiesFromSingleStereotypeTest.class)
```

Ladicek:
Otherwise LGTM.

impl/src/main/java/org/jboss/cdi/tck/tests/definition/stereotype/broken/multiplePriorities/ConflictingPrioritiesFromSingleStereotypeTest.java
manovotn: Fixed

Ladicek (approved):

PR #312, opened 2021-11-08
Initial TCK tests for build compatible extensions

Ladicek:


Ladicek:
Marked as draft, because I didn't run these tests with any implementation yet. They are, however, a direct port of tests I developed for my Quarkus prototype, so they should mostly be fine.

I'll work on running these tests with an implementation next.

Ladicek:
Amended the commit with a test for synthetic observers of parameterized types.

Ladicek:
Rebased.

As discussed on the CDI call today, we still don't have a way to run this with an implementation, but work is ongoing to run this both with Weld and Quarkus. Once that happens, this PR will be marked as ready for review and assuming no problems are found, merged. More tests need to be added for sure, but those won't block this PR.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/changeBeanQualifier/ChangeBeanQualifierTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.jboss.test.audit.annotations.SpecVersion;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertTrue;
+
+@SpecVersion(spec = "cdi", version = "4.0")
+public class ChangeBeanQualifierTest extends AbstractTest {
+    @Deployment
+    public static WebArchive createTestArchive() {
+        // no beans.xml + an extension = not a bean archive, bean classes are added through the extension
```
manovotn, line 16: This is actually wrong - the archive will have `beans.xml` automatically included (`WebArchiveBuilder` does that).
Instead, we could make use of the `annotated` discovery mode (just the empty beans.xml file) and simply not give those beans any BDA.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/changeBeanQualifier/MyOtherService.java
```diff
@@ -0,0 +1,10 @@
+package org.jboss.cdi.tck.tests.build.compatible.extensions.changeBeanQualifier;
+
+import jakarta.enterprise.context.Dependent;
+import org.jboss.arquillian.core.api.annotation.Inject;
+
+@Dependent
+public class MyOtherService {
```
manovotn, line 7: Your deployment assumes no discovery yet this bean is not added via extension.

manovotn (changes requested):
I have glued together several snapshots and I am able to test this with Weld.
Most of the tests are failing, so I am going to look into what's the cause and contribute the changes back here.

Meanwhile I dropped a few comments to things I noticed.

manovotn:
I have added a commit that fixes `ChangeBeanQualifierTest` to demonstrate the issue we described in https://github.com/eclipse-ee4j/cdi/issues/564
@graemerocher ^ just to make clear what we meant :)

manovotn:
Added another commit that fixes all but `SyntheticBeanTest` where I am not quite sure what the issue is yet.

manovotn:
Ok, added another commit that fixes the last test. With these changes I can get Weld's implementation to pass it.

manovotn:
Moved from draft to standard PR.
I will also give it green light from myself but maybe @Ladicek would like to review it before merging since I made quite a few changes :)

manovotn (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBean/SyntheticBeanExtension.java
```diff
@@ -30,7 +30,7 @@ public void synthesize(SyntheticComponents syn, Types types) {
                         .member("enumeration", MyEnum.class, "NO")
                         .member("type", types.ofClass("org.jboss.cdi.tck.tests.build.compatible.extensions.syntheticBean.MyEnum")
                                 .declaration())
-                        .member("nested", AnnotationBuilder.of(
+                        .member("nested", AnnotationBuilder.of( // TODO doesn't owrk
```
Ladicek, line 33: ??? :-)

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/customStereotype/MyService.java
```diff
@@ -1,5 +1,6 @@
 package org.jboss.cdi.tck.tests.build.compatible.extensions.customStereotype;
 
+@DummyStereotype
```
Ladicek, line 3: Why is this required? The class already has a stereotype?

manovotn: Originally, `MyService` was registered via `@Discovery` manually, I simply made sure it will be replaced by natural discovery process.

Ladicek:
Hi @manovotn, I've pushed a few more commits with some proposed changes, so that:

- the tests pass on ArC;
- multiple combinations of `beans.xml` present/missing and classes added/not added via `@Discovery` are tested;
- this still follows the rules for type discovery as we discussed.

If you can let me know how I can test with Weld, I'd be happy to do that.

manovotn:
@Ladicek I am seeing one failing test with Weld. `SyntheticObserverOfParameterizedTypeExtension` is adding `MyService` but that class also contains a bean defining annotation and as such is discovered. 

The way to execute it with Weld is:
* Check out master branch
* Perform a `mvn clean install -DskipTests`
* In `core/pom.xml`, change the value of `<cdi.tck-4-0.version>` property to `4.0.0-SNAPSHOT`
  * This OFC assumes you have CDI TCK built off this PR
* Run `mvn clean verify -f jboss-tck-runner/pom.xml`

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticObserverOfParameterizedType/SyntheticObserverOfParameterizedTypeExtension.java
```diff
@@ -0,0 +1,23 @@
...
+import jakarta.enterprise.inject.build.compatible.spi.Synthesis;
+import jakarta.enterprise.inject.build.compatible.spi.SyntheticComponents;
+import jakarta.enterprise.inject.build.compatible.spi.Types;
+
+import java.util.List;
+
+public class SyntheticObserverOfParameterizedTypeExtension implements BuildCompatibleExtension {
+    @Discovery
+    public void discovery(ScannedClasses scan) {
+        scan.add(MyService.class.getName());
```
manovotn, line 15: @Ladicek This is redundant and causes failures for Weld.
`MyService` has `@ApplicationScoped` and as such will be discovered.
Will you fix that or should I?

web/src/main/java/org/jboss/cdi/tck/shrinkwrap/ee/WebArchiveBuilder.java
```diff
@@ -94,7 +94,9 @@ protected WebArchive buildInternal() {
         processResources(webArchive);
 
         // Deployment descriptors
-        webArchive.addAsWebInfResource(getBeansDescriptorAsset(), buildBeansDescriptorTargetPath(getBeansDescriptorTarget()));
+        if (includeBeansXml) {
```
manovotn, line 97: Do we actually need to have two `WebArchiveBuilder` classes? There is one in `impl` and other under `web`; seems like a remnant of the TCK split.

starksm64: The one in in the web module has an ejb module that is included by default in test deployments and has a dependency on the EnterpriseArchiveBuilder.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticObserverOfParameterizedType/SyntheticObserverOfParameterizedTypeExtension.java
Ladicek: I would call that a bug _somewhere_. I intentionally added 1 or 2 cases where `@Discovery` is used to add a bean which is discovered through normal type discovery. I'm not aware of anything in the spec that would say this is illegal, so if that's what should happen in your opinion, we need to specify that somewhere.

web/src/main/java/org/jboss/cdi/tck/shrinkwrap/ee/WebArchiveBuilder.java
Ladicek: Perhaps so, I didn't really check.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticObserverOfParameterizedType/SyntheticObserverOfParameterizedTypeExtension.java
manovotn: @Ladicek Yes you can do that. The discovered and added types (based off the same java class) are two different ones but with equal bean types and qualifiers (unless changed later on in the bootstrap OFC).
You should be getting an ambiguous resolution exception.

Taking portable extensions as an example - if you take a look at BBD javadoc, it has a specific parameter that serves this purpose - https://github.com/eclipse-ee4j/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/BeforeBeanDiscovery.java#L162
In fact, you could in theory add one AT several times with just different IDs.

Ladicek: My expectation actually was that the container would ignore the duplicate :-) We need to specify this somehow.

Ladicek:
Added one commit to avoid the type discovery duplicate.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticObserverOfParameterizedType/SyntheticObserverOfParameterizedTypeExtension.java
Ladicek: This was specified in https://github.com/eclipse-ee4j/cdi/pull/571 (as non-portable) and the test was corrected, so resolving this open question.

Ladicek:
I think this is ready to be merged now, because it passes on ArC and Weld. I'll just restructure the commits a little and squash a bunch of them.

Ladicek:
Done.

graemerocher (approved):

manovotn (approved):

PR #319, opened 2021-12-06
Add test for build compatible extensions being able to see transformed annotations

Ladicek:


impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/seeChangedAnnotations/SeeChangedAnnotationsExtension.java
```diff
@@ -0,0 +1,58 @@
...
+    public void enhancement(ClassConfig clazz) {
+        clazz.addAnnotation(AnnotationBuilder.of(MyAnnotation.class).value(clazz.info().simpleName()).build());
+        for (MethodConfig method : clazz.methods()) {
+            method.addAnnotation(AnnotationBuilder.of(MyAnnotation.class).value(method.info().name()).build());
+        }
+    }
+
+    @Validation
+    public void validation(Messages msg, Types types) {
+        Type type = types.of(MyService.class);
```
manovotn, line 26: I don't think this makes much sense and you cannot expect it to be the same as the type modified during enhancement.

Firstly, validation phase happens as the last one. This means you have done type discovery (where `Types` matter) and moved on to creation of `Bean` objects. After that, I don't think types, or their inner states, matter much, if at all.

Secondly, in Full CDI (and with AT), creating a new AT off a java class results in a brand new AT object which isn't really surprising given that you can have several ATs based off the very same class (see `BBD.addAnnotatedType`). On top of that you can have synthetic AT implementing the interface itself and I don't see how you'd match it to that. I am stating this point mainly since we need integration with CDI Full and ATM that is solely via PE implementation.

Ladicek: `Types` is just a util class to create instances of `Type` and may be used in all phases after `@Discovery`. For what it's worth, this test could be in any phase after `@Enhancement`, and it could also be in `@Enhancement` itself. At least that's how I've always been thinking about this API.

Even if not, a `BeanInfo` has a method to obtain the declaring class. It is my firm opinion that the resulting `ClassInfo` just has to show what `@Enhancement` did. (Though it's probably not explicitly stated anywhere.)

The trouble with Portable Extensions is that there's no way how to consistently obtain the same `AnnotatedType` throughout the execution. You can create one with an ID, but you can't get to it later -- neither from `ProcessAnnotatedType`, nor from the `BeanManager`.

Ladicek: I've amended the test a bit to verify how `ClassInfo` for `MyService` looks like from 3 different entry points: `@Enhancement` (executed later than the one doing annotation transformation), `@Registration` (for the bean of type `MyService`), and `@Validation` (as before). These should all work in my opinion.

graemerocher (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/seeChangedAnnotations/SeeChangedAnnotationsExtension.java
manovotn: > The trouble with Portable Extensions is that there's no way how to consistently obtain the same AnnotatedType throughout the execution. 

Exactly. And I don't think it's worth implementing another abstraction layer in Full just to make this work for Lite extension integration.

This test is based on the premise that each java class has exactly one `Type` representation in CDI (or `AnnotatedType` in CDI Full). But this isn't true in Full and hence I don't see it being true in Lite either. Look at the case we discussed last week - having certain type discovered while also being artificially added via `@Discovery`. This scenario doesn't need to result in an override (it doesn't for CDI Full, anyway).
With that in mind, trying to create `Type` from any `Class<?>` logically has to create a new object instead of reffering to a cached state simply because you might have to "guess" which one to refer to.

Ladicek: > This test is based on the premise that each java class has exactly one `Type` representation in CDI (or `AnnotatedType` in CDI Full). But this isn't true in Full and hence I don't see it being true in Lite either.

It certainly is true in Lite. It's one of the basic assumptions the entire Build Compatible Extensions API makes. Some might see it as a regression compared to Portable Extensions, but it is most definitely what I had in mind the whole time.

manovotn: I can see why that is the case for Lite but I don't think we want to bake it into TCKs because it doesn't make much sense for any Full implementation supporting Lite extensions.
Full implementations can mix both extension models. Therefore, you can get into situation where there is more than one AT based off some `Foo.java` class - one discovered and one or more added via Portable extensions (`BBD.addAnnotatedType`). In such case, what do you expect to see if you then add Lite extension and perform `Types.ofClass(Foo.class)`?

Therefore, -1 for this test.

Ladicek:
After discussing this with @manovotn, I'm retracting this test, because while it verifies my original intent, it can't really work in CDI Full. Too bad :-/

PR #326, opened 2021-12-08
Review tests using @SessionScoped and move/change them if needed.

### LINKED ISSUE #325: Review if tests with @SessionScope are under Full category
If https://github.com/eclipse-ee4j/cdi/pull/577 we need to review all tests using this scope and whether they have Full category.
A quick `grep` shows a bunch of them, but not too many, which are possible candidates for this move.

manovotn:
Fixed #325 

Some tests are moved under `full` package, some are tweaked because, for instance, the scope used in the test doesn't matter so long as it's normal scope etc.

Tested with Weld embedded:
```
[INFO] Tests run: 1237, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 38.761 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1237, Failures: 0, Errors: 0, Skipped: 0
```


Ladicek:
Looks good, but there's a couple of newly added `*Test` classes in the `full` package that I assume were moved from non-`full`, but I don't see corresponding deletions. Maybe you forgot to add something to the commit?>

manovotn:
> Looks good, but there's a couple of newly added `*Test` classes in the `full` package that I assume were moved from non-`full`, but I don't see corresponding deletions. Maybe you forgot to add something to the commit?

I don't think I did.
Newly added tests keep using `@SessionScoped` beans while retained tests had some bits extracted or changed to use other scopes (mostly req. scope). An example of this would be `GetWithNoCreationalContextTest`.

Ladicek:
Ah I see! Sorry, I totally missed that. That's nice!

Ladicek (approved):

PR #328, opened 2021-12-08
Introduce simple test for OM#getDeclaringBean

### LINKED ISSUE #324: Provide test coverage for OM exposing Bean
Issue is described in https://github.com/eclipse-ee4j/cdi/issues/563
If we decide to accept the proposed solution, we'll need a simple TCK test that asserts this for standard class-based bean.

manovotn:
Fixes #324 

First commit is the test addition, second is a small correction to unrelated needed because of generics change in CDI API.

Creating as a draft because this PR will also need a bump in CDI API version.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/method/StockWatcher.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/method/StockWatcher.java
@@ -16,10 +16,14 @@
  */
 package org.jboss.cdi.tck.tests.event.observer.method;
 
+import jakarta.annotation.Priority;
 import jakarta.enterprise.context.Dependent;
 import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.inject.Alternative;
 
 @Dependent
+@Alternative
+@Priority(1)
 public class StockWatcher {
 
     private static Class<?> observerClazz;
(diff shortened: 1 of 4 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/method/StockWatcher.java
```diff
@@ -16,10 +16,14 @@
...
 package org.jboss.cdi.tck.tests.event.observer.method;
 
+import jakarta.annotation.Priority;
 import jakarta.enterprise.context.Dependent;
 import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.inject.Alternative;
 
 @Dependent
+@Alternative
+@Priority(1)
```
Ladicek, line 26: Are these needed? It was a selected bean before, no?

Ladicek: Ah, it's just to be able to assert more facts about the bean, right?

Ladicek:
Looks fine.

I didn't realize that making the `handles()` return type covariant would require such change, but I believe that's only for people who call `instance.handles().iterator()` directly. If you use foreach, it should be fine.

impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/method/StockWatcher.java
manovotn: Exactly, I added that solely to be able to assert a few things about the bean :-)

manovotn:
> I didn't realize that making the handles() return type covariant would require such change, but I believe that's only for people who call instance.handles().iterator() directly. If you use foreach, it should be fine.

Correct, using `for (Instance.Handle<Processor> p : instance.handles()) {...}` works just fine.

Ladicek (approved):

manovotn:
Updated CDI API.

PR #331, opened 2021-12-14
Change lang model TCKs to pass on both, JDK 11 and 17 if using reflec…

### LINKED ISSUE #330: Lang model TCK issues on newer JDKs
As part of https://github.com/eclipse-ee4j/cdi/issues/583 I encountered several parts on the lang model TCK that were problematic with JDK 16 while passing with JDK 11.

This concerns any runtime CDI Full implementation (in this case Weld) that will use Java reflection as a basis for the lang model.

An example of the issue is a method with receiver containing generics:
```
    public void methodWithReceiver(@AnnReceiver1 ReceiverOnGenericClass<T> this) {
    }
```

When accessed via reflections. with: `ReceiverOnGenericClass.class.getDeclaredMethod("methodWithReceiver").getAnnotatedReceiverType().getType()`

this will return:
* with JDK 11 - `{Class} "class org.jboss.cdi.lang.model.tck.ReceiverOnGenericClass"`
* with JDK 16 - `{ParameterizedTypeImpl} "org.jboss.cdi.lang.model.tck.ReceiverOnGenericClass<T>"`

This then leads to conflicting expectationd in the lang model interpretation and the TCK assertions.

I have found three cases which are problematic with different JDKs; I'll send a PR that changes the TCK to accept both variants with a `TODO` comment explaining it.
As a side note, none of them affects CDI functionalities.

manovotn:
…tion as basis for lang model.

Fixes #330 

Adapted the TCK to pass for both JDK versions that I tested with (11,17).

There is probably going to be more weirdness with other JDKs, but I haven't tested them all.
For instance I have noticed a discrepancy between JDK 11 and JDK 17 in how they handle annotated Enum constructor parameters (apparently, JDK 11 moves all your annotations over to a synthetic parameter??!). I am thinking maybe we should state what JDK is the lang model supposed to be tested with for purposes of compatibility (i.e. JDK 11 for EE 10) and then review for every major JDK release. That, or we need to make sure we try it with every JDK version that's out and keep iterating on it which might be expensive.

I'm open to suggestions.
Also CCing @Ladicek - would love to hear your thoughts on this once you're back.

### DIFF
--- a/lang-model/src/main/java/org/jboss/cdi/lang/model/tck/AnnotatedReceiverTypes.java
+++ b/lang-model/src/main/java/org/jboss/cdi/lang/model/tck/AnnotatedReceiverTypes.java
@@ -68,13 +68,18 @@ public class AnnotatedReceiverTypes<T> {
     private static void verifyMethodWithReceiverOnGenericClass(ClassInfo clazz) {
         MethodInfo method = LangModelUtils.singleMethod(clazz, "methodWithReceiver");
 
-        // reflection returns `AnnotatedType` which is not `AnnotatedParameterizedType`,
-        // so here, we expect the same -- `ClassType`, not `ParameterizedType`
         assert method.receiverType() != null;
-        assert method.receiverType().isClass();
-        assert method.receiverType().asClass().annotations().size() == 1;
-        assert method.receiverType().asClass().hasAnnotation(AnnReceiver1.class);
-        assert method.receiverType().asClass().declaration().name().equals("org.jboss.cdi.lang.model.tck.ReceiverOnGenericClass");
+        // Following assertions yield different results with JDK 11 and JDK 17 when accessed through reflection, TCK accepts both variants
+        assert method.receiverType().isClass() || method.receiverType().isParameterizedType();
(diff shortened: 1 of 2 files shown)


manovotn:
CI is failing because `master` branch depends on RC artifacts which are only staged ATM but not released.

Ladicek:
I vaguely recall scratching my head about the receiver type thing. I originally also thought it should be a parameterized type, but both reflection and Jandex gave me a class type, so I figured they probably know what they're doing :-)

I think change like this is fine, thanks for going through it!

starksm64:
Java SE 17 is a requirement for running the TCK, so that is what we need to test with and document as the language model we support. We really can only focus on the LTS releases of Java SE, so for EE10 that is SE 11 and 17.

manovotn:
I have reworded this PR to mention JDK 11 and 17 as basis for testing and have tested against these versions with Weld.
I've also changed all `TODO` notes to plain comments explaining the cause.

We'll need to get this fix into some released artifact - probably re-staging the current CR? CC @starksm64

Ladicek (approved):
LGTM

PR #333, opened 2022-01-10
Add TCK smoke test for coexistence of portable extensions and build c…

### LINKED ISSUE #332: Create smoke test for @SkipIfPortableExtensionPresent
Currently, we have no coverage for `@SkipIfPortableExtensionPresent` - an annotation that can be used when portable extensions and build compatible extensions co-exist in a CDI Full environment.

We should add a simple smoke test for it.
Note that this will have to belong to a `Full` category.

The way the test will look will be affected by resolution of https://github.com/eclipse-ee4j/cdi/issues/585

manovotn:
…ompatible extensions.

Fixes #332


manovotn:
On the meeting we agreed that we'll merge this so that we're able to stage another CR with this change included.

CC @starksm64

Issue #334, opened 2022-01-25
Verify that Lite tests are not using a beans.xml content other than bean-discovery-mode attribute to affect behavior


starksm64:
It was brought up on today's spec call that there are Lite tests that are dictating behavior by beans.xml content other than the bean-discovery-mode attribute. These need to be investigated.

manovotn:
> It was brought up on today's spec call that there are Lite tests...

Any chance we can ask whoever identified the issue to give us name of those tests? Going through all tests out there would be very time demanding :-)
I suppose those are some tests that manually add `beans.xml`? Since those added via `WebArchiveBuilder().withBeansXml()` should add anything changing the behavior.

starksm64:
The first one pointed out was org.jboss.cdi.tck.tests.alternative.broken.incorrect.name.stereotype.NoAnnotationWithSpecifiedNameTest that is expecting a failure due to a test specific beans.xml. It is those tests calling withBeansXml(...) to add test specific mods without a full group tag that are the problem. There do seem to be a few.

manovotn:
@starksm64 that test doesn't have wrong beans.xml, it should simply be a CDI Full test.
Its purpose is to verify what happens if you declare a non-existing stereotype annotation inside beans.xml.

starksm64:
Right, that is the problem; It is not in the full test group, so Micronaut was seeing this as a failure while running the Lite tests. Adding the correct group to these tests is the point of this issue.

manovotn:
Aah ok, I thought the issue is that there are actual tests aimed to test Lite behavior but have CDI Full-like beans.xml.
Got it.

PR #337, opened 2022-01-27
Put cdi-full group on tests that rely on non-empty beans.xml to affect behavior

starksm64:
#334


starksm64:
I did not change the package name of these tests to be under the org.jboss.cdi.tck.tests.full.* package. Is this something that should be done?

Ladicek:
In hindsight, I wish we never agreed to move the full-only tests into a `full` package, it's only made a mess. But we should be consistent, so I'd move these tests too :-)

I didn't look at the tests properly, perhaps some of them could/should be copied/rewritten to not use `beans.xml`, so that we don't lose coverage of Lite.

manovotn:
They should be moved under `full` package for consistency.

However, some of these tests seem like they could be just adjusted as they don't really test contents of `beans.xml`.
I'll look into it and provide a PR.

manovotn:
Closing in favor of https://github.com/eclipse-ee4j/cdi-tck/pull/338

Issue #335, opened 2022-01-25
Verify documentation around cdiLiteMode cdi-tck.properties flag


starksm64:
A miss named cdiLiteMode cdi-tck.properties flag was resulting in a datasource setting check failure. We are specifying Lite behavior via the cdiLiteMode flag and the test groups. Is this documented correctly and do we need two settings?

PR #339, opened 2022-01-28
Document the cdiLiteMode flag and add a core profile config example, …

starksm64:
…#335

### DIFF
--- a/doc/reference/src/main/asciidoc/configuration.asciidoc
+++ b/doc/reference/src/main/asciidoc/configuration.asciidoc
@@ -17,8 +17,7 @@ You should set the following required properties:
 [options="header"]
 |===============
 |Property = Example Value|Description
-|+org.jboss.cdi.tck.libraryDirectory=/path/to/extra/libraries
-                     +|
+|+org.jboss.cdi.tck.libraryDirectory=/path/to/extra/libraries+|
                      The directory containing extra JARs to be placed in
                      the test archive library directory such as the porting package
                      implementation.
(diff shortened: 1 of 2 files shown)


doc/reference/src/main/asciidoc/configuration.asciidoc
```diff
@@ -17,8 +17,7 @@ You should set the following required properties:
 [options="header"]
 |===============
 |Property = Example Value|Description
```
Ladicek, line 19: Maybe this table (or the one below?) should also have a line for the `cdiLiteMode` flag?

Ladicek:
Other than that, looks fine.

Ladicek (approved):

manovotn:
So what's the purpose of this property? Just to indicate that other configuration is not needed?

starksm64:
That is what it has become at this point. In previous iterations of the build time stuff we were also adding source, and triggering the annotation processor.

manovotn:
I knew we were. I am rather asking we if still need this configuration option for something.

starksm64:
It is annoying to have to set the various jdbc, jms properties if you are running under CDI Lite. Do you have an alternative suggestion for how to deal with this?

manovotn:
> It is annoying to have to set the various jdbc, jms properties if you are running under CDI Lite

Aah, so it is there to avoid the other configs. Ok, got it now.
Let's keep it, I just wasn't sure I understood it completely. Sorry for the confusion.

PR #341, opened 2022-02-03
Move remaining EL tests to cdi-full

starksm64:


starksm64:
Don't know why I could not get these commits squashed before the PR.

Ladicek:
I suppose there's just one commit with the actual relevant changes, and the rest is an artifact of branch management? If so, here's a branch that has just that one commit on top of current master: https://github.com/eclipse-ee4j/cdi-tck/compare/master...Ladicek:el-tests-to-full?expand=1

starksm64:
Yes, there is only one real commit here. I tried screwing around with squashing commits on another repo but could not get it cleaned up. Just create a PR from your branch and I'll close this one.

Ladicek:
Created #342, thanks!

PR #350, opened 2022-02-16
Move more tests using passivation and EL into Full test group.

### LINKED ISSUE #340: Beans, Contexts, and EL org.jboss.cdi.tck.spi integration interfaces
Need to look at tests that are using the Beans, Contexts, and EL org.jboss.cdi.tck.spi integration interfaces that are not currently in a cdi-full test group.

manovotn:
Fixes #340

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/full/context/dependent/DependentContextTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/full/context/dependent/DependentContextTest.java
@@ -12,6 +12,7 @@ import org.jboss.test.audit.annotations.SpecAssertion;
 import org.jboss.test.audit.annotations.SpecVersion;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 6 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/full/context/dependent/DependentContextTest.java
```diff
@@ -46,4 +47,14 @@ public void testDependentsDestroyedWhenElEvaluationCompletes() throws Exception
         assert FoxRun.isDestroyed();
         assert Fox.isDestroyed();
     }
+
+    @Test
+    @SpecAssertion(section = DEPENDENT_CONTEXT, id = "g")
+    // Dependent context is now always active
+    public void testContextIsActiveWhenEvaluatingElExpression() {
```
manovotn, line 54: TBF I am not sure this test is even worth having since I cannot imagine an implementation now having dependent context always active... but then again, what do I know? :-)

starksm64 (approved):

PR #352, opened 2022-03-17
Allow excluding type annotation checks

dstepanov:
There is a bug in the annotation processor not providing type annotations for classes loaded from the classpath.
https://bugs.openjdk.java.net/browse/JDK-8225377


graemerocher (approved):

Ladicek (approved):
Looks good to me, though I'd probably call the field `RUN_TYPE_ANNOTATION_TESTS` (defaulting to `true`) so that the `if` statements don't have to negate the condition. But that's just nitpicking.

dstepanov:
I did rename the property. Not sure why the build is failing.
It would be nice to have it released so we can integrate it into the test suite.

Ladicek:
The build fails most likely because of this commit: https://github.com/eclipse-ee4j/cdi-tck/commit/b557cc96f4c9c68a3f923a588231ff8bba29542c That was introduced in preparation of the final CDI 4.0 release I think. Let me check locally.

Ladicek:
I rebased this PR locally on the 4.0.0.RC6 tag and `mvn clean verify` passes just fine. So the build failures are just an artifact of preparation for staging the final build. That is, this PR may go in, in my opinion.

I don't know what's the process right now, but I believe we're retracting the 4.0.0 release due to a TCK deficiency anyway (see #353), so we should be able to get this in for the 4.0.0 release too. CC @starksm64

PR #354, opened 2022-03-18
[WIP] First step at adding Startup/Shutdown events

starksm64:
This is just one step to verify the Startup event. More will be coming.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/priority/contextLifecycleEvent/ApplicationContextLifecycleEventObserverOrderingTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/priority/contextLifecycleEvent/ApplicationContextLifecycleEventObserverOrderingTest.java
@@ -41,7 +41,7 @@ public class ApplicationContextLifecycleEventObserverOrderingTest extends Abstra
     @SpecAssertion(section = OBSERVER_ORDERING, id = "b")
     public void testContextLifecycleEventOrdering() {
         // ActionSequence was originally used, but we cannot reliably reset it without extensions
-        String expectedString = "ABCD";
+        String expectedString = "ABCDSu";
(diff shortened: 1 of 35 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/priority/contextLifecycleEvent/ApplicationContextLifecycleEventObserverOrderingTest.java
```diff
@@ -41,7 +41,7 @@ public static WebArchive createTestArchive() {
     @SpecAssertion(section = OBSERVER_ORDERING, id = "b")
     public void testContextLifecycleEventOrdering() {
         // ActionSequence was originally used, but we cannot reliably reset it without extensions
-        String expectedString = "ABCD";
+        String expectedString = "ABCDSu";
```
manovotn, line 44: We cannot test it like this - there is nothing in the specification guaranteeing ordering of events. Namely that of `@Initiliazed` for app scope and the new `Startup`. Similar situation is for `@Destroyed` and `Shutdown`.

It will either have to b a separate test that assert only that these events occurred or a test with both but disregarding their ordering.

manovotn: I could come up with something on Mon, I will likely have to write a test for Weld anyway so I might as well just adapt it for TCKs while at it.

manovotn: Actually, I was wrong - we can assert *certain* ordering. I forgot we worded it so that we require the `@Initialized` event for app scope to be fired before `Startup` and likewise to fire `Shutdown` before we do `BeforeDestroyed` for app scope.

PR #355, opened 2022-03-22
Introduce tests for Startup/Shutdown, revert versions

### LINKED ISSUE #353: Create tests to verify that Startup and Shutdown events are fired
I can not find any CDI 4.0.0 TCK tests that actually validate that the new Startup and Shutdown events are correctly fired.
The events were added to the API and spec here: https://github.com/eclipse-ee4j/cdi/pull/543

At the moment in 4.0.0, there is a single reference to Startup in a BCE test but I do not think that test checks the event was fired, only that an observer is present
https://github.com/eclipse-ee4j/cdi-tck/blob/d78a11a2dabac738f9bfa1bc3a82a6ae1f3df52c/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/registration/MyServiceFoo.java#L14

manovotn:
Fixes #353 

Similar to tests I've added to Weld.
In classic test, we are unable to verify `Shutdown` event but there is also an SE test where both can be verified just fine.

I have also added a commit that reverts versions back to latest released so that we can have a pass in CI and release new RC (which I'll do, likely tomorrow).

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/event/lifecycle/ObservingBean.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/event/lifecycle/ObservingBean.java
@@ -1,0 +1,34 @@
+package org.jboss.cdi.tck.tests.event.lifecycle;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.context.BeforeDestroyed;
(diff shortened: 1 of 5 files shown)


Ladicek (approved):
LGTM

PR #360, opened 2022-03-24
Make BCE invalid tests include extension class

### LINKED ISSUE #359: Invalid Build Compatible Extension tests don't include the extension class
Tests for invalid Build Compatible Extension tests (e.g. [EnhancementMultipleParamsTest](https://github.com/eclipse-ee4j/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/invalid/EnhancementMultipleParamsTest.java)) are only adding the service provider file to the test archive and are not adding the actual extension class.

Depending on how the test is run against a given container, the class may be found anyway (if all test code is on the classpath) or the container may throw a `DeploymentException` anyway and pass the test. They fail for us on Open Liberty with a `ClassNotFoundException` when trying to load the extension class.

Azquelt:
Previously, the test was adding the service provider file but not the
actual extension class.

Fixes #359


Ladicek (approved):
Looks good to me. I don't think this was intentional, was it @manovotn?

manovotn (approved):
Nope this wasn't intentional and the PR looks good. Thanks @Azquelt!
Interestingly, these tests work just fine for Weld embedded and I am getting correct results there.
I guess that's because of some underlying CL differences.

PR #362, opened 2022-03-24
IntrospectServlet needs to be a bean for JEEComponentTest

### LINKED ISSUE #361: IntrospectServlet in JavaEEComponentsTest should be a bean
I think this class would have been a bean until we changed the default discovery mode to annotated in #250.

Now we either need to add an annotation to `IntrospectServlet` or set the bean discovery mode back to all in this test.

Azquelt:
This used to be a bean because we used discovery mode all. However, the
default discovery mode has changed in 4.0 to be annotated, so the class
needs an annotation to be a bean.

Fixes #361

### DIFF
--- a/web/src/main/java/org/jboss/cdi/tck/tests/implementation/simple/ee/components/IntrospectServlet.java
+++ b/web/src/main/java/org/jboss/cdi/tck/tests/implementation/simple/ee/components/IntrospectServlet.java
@@ -44,6 +44,7 @@ import org.jboss.cdi.tck.util.ActionSequence;
  */
 @SuppressWarnings("serial")
 @WebServlet("/test")
+@Dependent
 public class IntrospectServlet extends HttpServlet {
 
     public static final String MODE_INJECT = "inject";


manovotn:
The test can simply have `beans.xml` with discovery mode all to keep it at exactly previous behavior.

But that aside, does servlet need a bean defining annotation to be picked up? I thought servlets, filters (and at least one other thing I cannot recall) are just guaranteed to be injected into? And that alone is probably is defined in servlet spec?
I only found https://jakarta.ee/specifications/servlet/5.0/jakarta-servlet-spec-5.0.html#contexts-and-dependency-injection-for-jakarta-ee-platform-requirements
Which nicely refers to non-existent chapter :-)

Emily-Jiang:
> The test can simply have `beans.xml` with discovery mode all to keep it at exactly previous behavior.
> 
> But that aside, does servlet need a bean defining annotation to be picked up? I thought servlets, filters (and at least one other thing I cannot recall) are just guaranteed to be injected into? And that alone is probably is defined in servlet spec? I only found https://jakarta.ee/specifications/servlet/5.0/jakarta-servlet-spec-5.0.html#contexts-and-dependency-injection-for-jakarta-ee-platform-requirements Which nicely refers to non-existent chapter :-)

Servelet classes do not need bean defining annotations and they need to support injection, which is specified in the platform spec (see [here](https://jakarta.ee/specifications/platform/9.1/jakarta-platform-spec-9.1.html#a651))

Azquelt:
Its true that servlets (and other components) need to support injection, but `IntrospecServlet` also has a producer method, and those need to be on a managed bean.

Emily-Jiang:
> Its true that servlets (and other components) need to support injection, but `IntrospecServlet` also has a producer method, and those need to be on a managed bean.

ah, that makes sense. Yes, the producer method should be on a bean but servlet is not a bean without a bean defining annotation.

Azquelt:
> The test can simply have beans.xml with discovery mode all to keep it at exactly previous behavior.

I'm happy either way. The requirement being tested is this:

> The instance used by the container to service an invocation of a Jakarta EE component will not be the same instance obtained when using `@Inject`, instantiated by the container to invoke a producer method, observer method or disposer method, or instantiated by the container to access the value of a producer field.

so I think the test is valid as long as the servlet is a bean (of any scope)

manovotn (approved):

PR #366, opened 2022-04-07
Add ability to trigger CI build action manually

starksm64:
Also add build using staging repo to validate staged dependencies
Update upper JDK bound to 17

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -1,6 +1,7 @@
 name: Jakarta Contexts and Dependency Injection TCK CI
 
 on:
+  workflow_dispatch:
   pull_request:
     branches: [ master ]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-actions.yml
```diff
@@ -11,13 +12,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [11, 16]
+        java: [11, 17]
```
Ladicek, line 15: Just for the record, `[11, 17]` is not an interval, it's an alternative syntax for lists. It is equivalent to this:

```yaml
jobs:
  build:
    name: "..."
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        java:
        - 11
        - 17
...
```

PR #368, opened 2022-04-08
Revert change to InvalidatingSessionDestroysConversationTest

### LINKED ISSUE #367: InvalidatingSessionDestroysConversationTest doesn't configure Faces correctly
It looks like [`InvalidatingSessionDestroysConversationTest`](https://github.com/eclipse-ee4j/cdi-tck/blob/master/web/src/main/java/org/jboss/cdi/tck/tests/context/conversation/InvalidatingSessionDestroysConversationTest.java) is meant to work like this:

* Make a request which is handled by Faces using `clouds.xhtml`, which looks up the ConversationScoped `Cloud` bean and starts a long-running conversation
* Assert that no `Cloud` bean has been destroyed yet (i.e. the conversation has not been destroyed yet)
* Invalidate the session
* Assert that a `Cloud` bean has been destroyed (because invalidating the session destroyed the conversation)

However, I don't think the test is set up correctly for Faces to handle the call to `clouds.xhtml`
* [The test makes a request to `clouds.xhtml`](https://github.com/eclipse-ee4j/cdi-tck/blob/master/web/src/main/java/org/jboss/cdi/tck/tests/context/conversation/InvalidatingSessionDestroysConversationTest.java#L63)
* However, the [web.xml only seems to configure the Faces servlet to handle requests matching `*.jsf`](https://github.com/eclipse-ee4j/cdi-tck/blob/master/web/src/main/resources/org/jboss/cdi/tck/tests/context/conversation/web.xml#L28-L31)
* This would mean that clouds.xhtml is just served as a plain file and the `Cloud` bean instance is never created and so isn't destroyed

It looks like the [3.0 version of this test](https://github.com/eclipse-ee4j/cdi-tck/blob/3.0/impl/src/main/java/org/jboss/cdi/tck/tests/context/conversation/InvalidatingSessionDestroysConversationTest.java#L60) made the request to `clouds.jsf` rather than `clouds.xhtml`. This was changed in #315 and I think it should be changed back.

Azquelt:
Request clouds.jsf rather than clouds.xhtml so that the request matches
the pattern for the faces servlet.

With this change, I have this test passing on Open Liberty.

Fixes #367

### DIFF
--- a/web/src/main/java/org/jboss/cdi/tck/tests/context/conversation/InvalidatingSessionDestroysConversationTest.java
+++ b/web/src/main/java/org/jboss/cdi/tck/tests/context/conversation/InvalidatingSessionDestroysConversationTest.java
@@ -60,7 +60,7 @@ public class InvalidatingSessionDestroysConversationTest extends AbstractConvers
         WebClient webClient = new WebClient();
 
         resetCloud(webClient);
-        webClient.getPage(getPath("clouds.xhtml"));
+        webClient.getPage(getPath("clouds.jsf"));
 
         assertFalse(isCloudDestroyed(webClient));
         invalidateSession(webClient);


manovotn:
Looks good to me but I hesitate to provide a binding review as I've also had troubles understanding this test.

manovotn (approved):
Actually, I cobbled together workable WFLY for this and I have the test passing with this change.
That makes two servers pass which, I think, gives us enough confidence to accept this.

PR #370, opened 2022-04-08
Set discovery-mode=all for non.contextual.ContainerEventTest

### LINKED ISSUE #369: non.contextual.ContainerEventTest requires discovery-mode=all
`org.jboss.cdi.tck.tests.lookup.injection.non.contextual.ContainerEventTest` checks that `ProcessAnnotatedType` events are fired for EE component objects (Servlet, Tag Handler, Tag Library, Filter, Session bean, JSF managed bean).

This will only happen if the component classes are also beans. In the 3.0 TCK, we defaulted to adding a beans.xml with discovery-mode=all, but in 4.0 we add a beans.xml with discovery-mode=annotated (changed in #250).

For this test to pass, either the test components need to have a bean defining annotation added, or a beans.xml must be added with bean-discovery-mode=all.

Azquelt:
This test relies on unannotated classes being discovered and so needs
discovery-mode=all.

This now needs to be added explicitly because the default was changed to
discovery-mode=annotated.

Also update the BeansXml javadoc with the new default discovery mode.

Fixes #369

### DIFF
--- a/impl/src/main/java/org/jboss/shrinkwrap/impl/BeansXml.java
+++ b/impl/src/main/java/org/jboss/shrinkwrap/impl/BeansXml.java
@@ -13,7 +13,7 @@ import java.util.List;
 
 /**
  * Allows to create a beans.xml file programmatically.
- * By default it uses discovery mode all and has version 3.0
+ * By default it uses discovery mode annotated and has version 3.0
  */
 public class BeansXml implements Asset {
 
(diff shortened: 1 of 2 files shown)


manovotn (approved):

manovotn:
The CI failures were due to dep on staged but unreleased Final artifacts and hence non-blocking.

PR #371, opened 2022-04-08
withDefaultPersistenceXml should use deployment time config

Azquelt:
`withDefaultPersistenceXml` is called in the deployment phase (to build the test application archive) so it should use the deployment time config.

Without this, the TCK expects the porting package and all its dependencies to be present on the test classpath, rather than just on the server. This made it much harder for me to run `ContainerEventTest` on my workstation since the porting package depends on server internals (and I don't think this test needs the porting package anyway).

Although I think this is a bug, it's been present in previous releases so everyone who has passed the older TCKs has clearly managed to work around it so we don't need to rush it in for 4.0.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/shrinkwrap/ArchiveBuilder.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/shrinkwrap/ArchiveBuilder.java
@@ -563,7 +563,7 @@ public abstract class ArchiveBuilder<T extends ArchiveBuilder<T, A>, A extends A
      */
     public T withDefaultPersistenceXml() {
         return withPersistenceXml(Descriptors.create(PersistenceDescriptor.class).createPersistenceUnit().name("test")
-                .jtaDataSource(ConfigurationFactory.get().getTestDataSource()).up());
+                .jtaDataSource(ConfigurationFactory.get(true).getTestDataSource()).up());
     }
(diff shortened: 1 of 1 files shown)


manovotn:
I suppose you mean `org.jboss.cdi.tck.tests.lookup.injection.non.contextual.ContainerEventTest` (there are 4 `ContainerEventTest` ATM).

Frankly speaking I have no idea how is this config supposed to work and I am unable to currently verify this against WFLY due to other issues so I cannot really provide a review.

manovotn:
@Ladicek actually, this is somewhat related to what was brought up in https://github.com/eclipse-ee4j/cdi-tck/issues/222
It is a usage of [Shrinkwrap Descriptors](https://github.com/shrinkwrap/descriptors), a project that wasn't updated in a while and the files it generates can be fairly outdated (the version of them at least).

I recall back then I was thinking if we shouldn't replace these usages with some other APIs of our own or update the project (which might be costly for just us). Or just plain replace them by XML files in project that could at least be updated by us.

manovotn:
> ... and I am unable to currently verify this against WFLY due to other issues so I cannot really provide a review.

BTW, I went through some loops to get an ad-hod working WFLY for this and I see no difference with and without this change.
I guess this can also be attributed to some differences in Arq. adapters?

Issue #378, opened 2022-06-03
Userguide is inconsistent with jboss-tck-runner examples in CDI TCK dist


starksm64:
The userguide shows the following setup for the Web Profile:

```
      <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-surefire-plugin</artifactId>
            <configuration>
                <excludedGroups>javaee-full</excludedGroups>
            </configuration>
       </plugin>
```

and no example setup for the Full Platform. However, the jboss-tck-runner/pom.xml setup uses the following for the Web Profile:
```
      <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-surefire-plugin</artifactId>
            <configuration>
                <excludedGroups>javaee-full,se</excludedGroups>
            </configuration>
       </plugin>
```

and the following for the Full Platform:
```
      <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-surefire-plugin</artifactId>
            <configuration>
                <excludedGroups>se</excludedGroups>
            </configuration>
       </plugin>
```

It really makes no sense to require a Jakarta EE server to support the Java SE CD APIs as it is a completely separate launch mode that would require a separate Arquillian container configuration. The userguide needs to be consistent with the WildFly/Weld jboss-tck-runner example.

starksm64:
This was updated in the 4.0.3 release.

PR #379, opened 2022-06-03
Make the configuration section explicit about all run modes

starksm64:
Addresses #378

### DIFF
--- a/doc/reference/src/main/asciidoc/configuration.asciidoc
+++ b/doc/reference/src/main/asciidoc/configuration.asciidoc
@@ -2,16 +2,16 @@
 
 == Configuration
 
-This chapter lays out how to configure the TCK Harness by specifying the SPI implementation classes, defining the target container connection information, and various other switches. You then learn how to setup a TCK runner project that executes the TCK test suite, putting these settings into practice. 
+This chapter lays out how to configure the TCK Harness by specifying the SPI implementation classes, defining the target container connection information, and various other switches. You then learn how to setup a TCK runner project that executes the TCK test suite, putting these settings into practice.
 
 [[tck-properties]]
 
 
 === TCK Properties
 
-System properties and/or the resource META-INF/cdi-tck.properties, a Java properties file, are used to configure the TCK. 
+System properties and/or the resource META-INF/cdi-tck.properties, a Java properties file, are used to configure the TCK.
 
-You should set the following required properties: 
+You should set the following required properties:
 
 .Required TCK Configuration Properties
 [options="header"]
@@ -25,16 +25,16 @@ You should set the following required properties:
 |+org.jboss.cdi.tck.testDataSource=java:jboss/datasources/ExampleDS+|
                      A few TCK tests work with Jakarta Persistence services and require a data source to be provided.
                      This property defines JNDI name of such resource. Required for the tests within the _persistence_ test group.
-                  
+
 |+org.jboss.cdi.tck.testJmsConnectionFactory=java:/ConnectionFactory+|
                     The JNDI name of the JMS test ConnectionFactory. Required for the tests within the _jms_ test group.
-                  
+
 |+org.jboss.cdi.tck.testJmsQueue=java:/queue/test+|
                     The JNDI name of the JMS test Queue. Required for the tests within the _jms_ test group.
-                  
+
 |+org.jboss.cdi.tck.testJmsTopic=java:/topic/test+|
                     The JNDI name of the JMS test Topic. Required for the tests within the _jms_ test group.
-                  
+
 
 |===============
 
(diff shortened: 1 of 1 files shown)


doc/reference/src/main/asciidoc/configuration.asciidoc
```diff
@@ -93,43 +90,87 @@ You can configure the CDI TCK to just run tests related to the CDI Lite specific
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
-                <excludedGroups>cdi-full,integration,javaee-full,se</excludedGroups>
+              <excludedGroups>cdi-full,integration,javaee-full,se</excludedGroups>
```
Ladicek, line 93: Since you only scan the `jakarta.enterprise:cdi-tck-core-impl` dependency, there should be no `integration` and `javaee-full` tests. Excluding just `cdi-full` and `se` should be enough.

doc/reference/src/main/asciidoc/configuration.asciidoc
```diff
@@ -93,43 +90,87 @@ You can configure the CDI TCK to just run tests related to the CDI Lite specific
...
+
+You can configure the CDI TCK to just run tests appropriate to the Java SE runtime by including the TestNG group _se_ and _arquillian_, e.g. via a maven-surefire-plugin configuration like:
+
+[source.XML, xml]
+----
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <groups>se,arquillian</groups>
```
Ladicek, line 149: Is there an `arquillian` group?

doc/reference/src/main/asciidoc/configuration.asciidoc
Ladicek: Actually it seems there are some `integration` tests in the `cdi-tck-core-impl` module... I'll need to look at it properly next week.

doc/reference/src/main/asciidoc/configuration.asciidoc
starksm64: Yes, it is used on the Arquillian class that does the TestNG integration. I copied this old issue over from Jira about the problem; https://github.com/arquillian/arquillian-core/issues/395. Until that is fixed, you need to specify it when stating which groups to run vs which groups to exclude.

doc/reference/src/main/asciidoc/configuration.asciidoc
starksm64: The org.jboss.cdi.tck.tests.event.observer.ObserverNotificationTest is marked as INTEGRATION, but this seems to only test features within the scope of Lite. Same with ConditionalObserverTest#testConditionalObserverMethodNotInvokedIfNoActiveContext. Yes, javaee-full should never appear in the cdi-tck-core-impl artifact.

manovotn: `cdi-tck-core-impl` module doesn't seem to have any tests with `javaee-full` group so we can perhaps remove any mention of that group from here @starksm64?

The `integration` group is technically ok to be present in `cdi-tck-core-impl` since that group states:

> Integration test - cannot run in embedded arquillian container

I take it that the embedded container is referring to [Weld's Arq. container](https://github.com/arquillian/arquillian-container-weld) which has some limitations due to its implementation (such as merging beans.xml or automatically activating contexts for test duration).
So any other CDI implementation, be it Lite of Full, can theoretically execute these tests fine (just like Weld can while running WFLY for instance).

manovotn (approved):
LGTM

breakponchito:
I'm seeing that the config example for the Core profile is excluding the following: `<excludedGroups>cdi-full,se</excludedGroups>`   While the text explaining the config say something different:  **_You can configure the CDI TCK to just run tests related to the CDI Lite specification appropriate for the Jakarta EE Core Profile by excluding TestNG groups cdi-full, integration, javaee-full, se, e.g. via a maven-surefire-plugin configuration like_**  is this correct?

starksm64:
Yes, and the core profile TCK includes an appropriate TestNG suite file that already has this configured, so you could just run using the suite.xml without any excluded group specified:

https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/core-profile-tck/cdi-tck-suite/src/main/resources/cdi-lite-tck-suite.xml

doc/reference/src/main/asciidoc/configuration.asciidoc
Ladicek: Ah OK, thanks, I didn't know that!

Ladicek:
I think @breakponchito highlights that there's an inconsistency between the text and the XML snippet. Which is fair I think, we should update the text.

doc/reference/src/main/asciidoc/configuration.asciidoc
```diff
@@ -93,43 +90,87 @@ You can configure the CDI TCK to just run tests related to the CDI Lite specific
...
+              <dependenciesToScan>
+                <dependency>jakarta.enterprise:cdi-tck-core-impl</dependency>
+            </configuration>
```
Emily-Jiang, line 152: This snippet misses one line to close the element:
</dependenciesToScan>
before the line 152
</configuration>

PR #385, opened 2022-06-20
Validation extension test should fail on `DefinitionException` not `DeploymentException`

dstepanov:
My understanding is that the compile-time errors tests should fail with `DefinitionException` and runtime/deploy errors with `DeploymentException`. Because extensions are executed at the compile-time it should fail on `DefinitionException`.

FYI @graemerocher

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/validation/ValidationTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/validation/ValidationTest.java
@@ -1,5 +1,6 @@
 package org.jboss.cdi.tck.tests.build.compatible.extensions.validation;
 
+import jakarta.enterprise.inject.spi.DefinitionException;
 import jakarta.enterprise.inject.spi.DeploymentException;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.container.test.api.ShouldThrowException;
@@ -14,7 +15,7 @@ import org.testng.annotations.Test;
 @SpecVersion(spec = "cdi", version = "4.0")
 public class ValidationTest extends AbstractTest {
     @Deployment
-    @ShouldThrowException(DeploymentException.class)
+    @ShouldThrowException(DefinitionException.class)
     public static WebArchive createTestArchive() {
         return new WebArchiveBuilder()
                 .withTestClassPackage(ValidationTest.class)


Ladicek:
My understanding, per https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#exceptions, is that definition errors are "local" or "intra-component" (e.g. related to a single bean), while deployment problems are "global" or "inter-component" (e.g. related to dependencies between beans). I don't think we considered updating the specification of definition errors/deployment problems to account for the difference between compile-time and runtime.

If I remember correctly, the build compatible extensions API almost always prescribes that errors lead to deployment problems. That might have been pure laziness on my side for sure :-)

dstepanov:
Ah I see, the only solution would be to prefix the error message otherwise there is no way to distinguish the type of the exception.

manovotn:
> I don't think we considered updating the specification of definition errors/deployment problems to account for the difference between compile-time and runtime.

+1
Basically, we didn't update it and before CDI Lite, there was no build and run time distinction.
I don't see much issue in expecting `DeploymentException` for what we are able to now perform in build time phase of frameworks.

Looking at the [javadoc](https://github.com/jakartaee/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/DeploymentException.java#L20-L42), I am not seeing any runtime slang that would prevent the usage:
> Deployment problems are detected by the container at initialization time. If a deployment problem exists in a deployment,  initialization will be aborted by the container.

And on top of that:
>  An implementation is permitted to throw a subclass of {@link DeploymentException} for any deployment problem.

This basically means MN can create any subclass they deem appropriate and throw that instead if it is an error you find during build time.

dstepanov:
In the reality, there is going to be a compilation error, so no exceptions, etc. For the testing framework, I need to come up with a way to throw an appropriate exception to satisfy `@ShouldThrowException`. The only solution I can come up with is to have a text error message produced by Javac prefixed by something and that can be a deciding factor on which exception to throw, that might not be something that we want to show to the user. Or the other solution is to change `ShouldThrowException` to catch `RuntimeException`.

Ladicek:
I guess you could formulate the error messages in a way that indicates whether a definition error or deployment problem occurred. I don't know how your error messages look like at the moment, but they could e.g. look like this:

```
DEFINITION ERROR
================

Producer method `com.example.MyBean#myProducer` has an `@Observes` parameter.
```

or

```
DEPLOYMENT PROBLEM
==================

Unsatisfied dependency at `com.example.MyBean#someField`.
```

But if you don't want to put this weird distinction in front of your users, you could hide it behind a flag. I can't think of a better solution at the moment.

manovotn:
This has been open for a while with no update and my understanding of the above conversation is that it is not really a bug/mistake.
Hence I am closing this PR, feel free to reopen if you think it is still unresolved.

PR #387, opened 2022-06-20
Improve matching custom `java.lang.annotation.Annotation`

### LINKED ISSUE #270: `java.lang.reflect.Type`s equality should't be asserted by `equals` for some cases
`org.jboss.cdi.tck.tests.event.metadata.EventMetadataTest`

Micronaut will return a custom instance of `ParameterizedType` which is not going to match the reflection implementation.

dstepanov:
Fixes #270

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/util/Assert.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/util/Assert.java
@@ -36,18 +36,6 @@ public class Assert {
     }
 
     /**
-     * Validate that types match.
-     *
-     * @param type The provided type
-     * @param requiredType The expected type
-     */
-    public static void assertTypeEquals(Type type, Type requiredType) {
-        if (!isTypesEqual(type, requiredType)) {
-            fail(String.format("Type '%s' does not match type '%s'", type, requiredType));
-        }
-    }
-
-    /**
      *
      * @param annotations
      * @param requiredAnnotationTypes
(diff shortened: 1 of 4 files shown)


impl/src/main/java/org/jboss/cdi/tck/util/Assert.java
```diff
@@ -96,9 +108,49 @@ public static void assertTypeListMatches(List<? extends Type> types, Type... req
...
+        for (Type existingType : existingTypes) {
+            if (isTypesEqual(requiredType, existingType)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static Boolean isTypesEqual(Type type, Type requiredType) {
+        return type.getTypeName().equals(requiredType.getTypeName());
```
Ladicek, line 141: This seems rather imprecise? Also `Type.getTypeName()` by defaults delegates to `toString()`, so it's also quite fragile in presence of custom implementations of `Type`. IIUC, the original code relied on `Type.equals()`, which is likely to be much better. Of course it requires custom implementations of `Type` to interoperate correctly, which isn't that hard to do AFAICT. Here are some custom implementations of `Type` that (should :-) ) work correctly: https://github.com/weld/core/tree/master/weld-lite-extension-translator/src/main/java/org/jboss/weld/lite/extension/translator/util/reflection

Ladicek: (It also requires the JDK implementations of `Type` to interoperate correctly, which AFAIK they do.)

dstepanov: You are right, I wound a bug in our type implementation. Keeping only annotations changes.

manovotn (dismissed):
I am fine with these changes but leaving the merge to @Ladicek as he had some comments to be addressed.

Ladicek:
No, my previous comment was addressed. But I keep wondering why this is necessary. Are you perhaps generating annotation classes that don't implement `annotationType()` properly?

dstepanov:
Good question, I will investigate more :)

manovotn:
@dstepanov is there any update on this PR? Should we keep it open?

dstepanov:
Not sure if it is still needed, we can close it for now

PR #388, opened 2022-06-20
Disable the tck-audit profile by default so that the org.jboss.test-a…

starksm64:
…udit dependencies are not seen by default in the TCK test artifact dependencies.

The build needs -DgenerateSource to run with this change

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,28 @@
 
 Check out the [TCK Reference Guide](https://eclipse-ee4j.github.io/cdi-tck/) to get acquainted with the CDI TCK and learn how to execute and debug it.
 
+## Building CDI TCK artifacts
+To build the CDI TCK artifacts, use:
+
+`mvn -DgenerateSource install`
+
+or when compiling against staged Jakarta artifacts:
+
+`mvn -DgenerateSource -Pstaging install`
+
+## Building the CDI TCK distribution
+The CDI TCK distribution artifact is built by specifing an additional `-Drelease` property to build the TCK reference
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -20,3 +39,4 @@ Master branch contains the work-in-progress on CDI TCK 4.0
 * lang-model - A standalone test suite for the CDI language model; see its [README](./lang-model/README.adoc)
 * web - The extra tests that depend on the web profile and full platform
 * README.md - this doc
+* settings.xml -
```
manovotn, line 42: There is no `settings.xml`?

impl/pom.xml
```diff
@@ -219,7 +219,7 @@
             <id>tck-audit</id>
             <activation>
                 <property>
-                    <name>!skipTckAudit</name>
+                    <name>generateSource</name>
```
manovotn, line 222: Since it was named audit, we could just keep it as `tckAudit` or something similar.

starksm64: Ok,  updated and added comment about why it is needed but not active by default

README.md
```diff
@@ -3,9 +3,28 @@
 
 Check out the [TCK Reference Guide](https://eclipse-ee4j.github.io/cdi-tck/) to get acquainted with the CDI TCK and learn how to execute and debug it.
 
+## Building CDI TCK artifacts
+To build the CDI TCK artifacts, use:
+
+`mvn -DgenerateSource install`
```
manovotn, line 9:
```suggestion
`mvn -DtckAuditSourceGen install`
```

PR #394, opened 2022-07-10
Fixed example Wildfly project for running the lang model TCK

OndroMih:
The example was designed to run with a plain Weld in Java SE. I added an option to run it against WildFly and fixed the test to pass on WildFly with similar fixes we applied for GlassFish.

I also updated the README to explain how to run the Language Model TCK.


dist-build/lang-model-tck-runner/src/test/java/org/jboss/weld/lang/model/tck/LangModelTckTest.java
```diff
@@ -51,8 +52,11 @@ public static Archive<?> deploy() {
                 // beans.xml with discovery mode "all"
                 .addAsWebInfResource(new BeansXml(BeanDiscoveryMode.ALL), "beans.xml")
                 .addAsServiceProvider(BuildCompatibleExtension.class, LangModelExtension.class)
-                // add this class into the deployment so that it's subject to discovery
-                .addClasses(LangModelVerifier.class);
+                .addClasses(LangModelExtension.class)
+                // add this package into the deployment so that it's subject to discovery, including its dependencies
+                .addPackage(LangModelVerifier.class.getPackage())
```
manovotn, line 57: If you add the package, the line `.addClasses(LangModelExtension.class)` is superflous. Or am I missing something?

dist-build/lang-model-tck-runner/src/test/java/org/jboss/weld/lang/model/tck/BuildtimeVetoExtension.java
```diff
@@ -0,0 +1,34 @@
...
+import jakarta.annotation.Priority;
+import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.inject.spi.Extension;
+import jakarta.enterprise.inject.spi.ProcessAnnotatedType;
+
+/**
+ * Vetos all classes for the build-time Lang Model TCK extension to avoid deployment issues with unaccessible package private members.
+ * @author Ondro Mihalyi
+ */
+public class BuildtimeVetoExtension implements Extension {
```
manovotn, line 27: A cleaner solution might be to change `LangModelTckTest` to use `annotated` discovery instead of `all`?
Another alternative is to change the TCK package and add `package-info.java` with `@Vetoed` in it.
Here is an example of what I mean from Weld code - https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/bean/package-info.java

manovotn (commented):
The `wildfly8` in the name of package should be changed to just `wildfly` (FTR the actual version will be 27 but it is just better without version for future iterations).

Also I didn't test the setup locally but I assume you did.

dist-build/lang-model-tck-runner/src/test/java/org/jboss/weld/lang/model/tck/LangModelTckTest.java
OndroMih: Both are necessary, as they are in different packages: `org.jboss.weld.lang.model.tck.LangModelExtension` while `org.jboss.cdi.lang.model.tck.LangModelVerifier`

manovotn: Oh my bad, I thought its fully copied from Weld where the [package is the same](https://github.com/weld/core/tree/master/lang-model-tck/src/test/java/org/jboss/weld/lang/model/tck) but it is only partial. Maybe we should move them to a shared package that doesn't contain `weld` in it?

manovotn: Sorry, my bad, misread the configuration.

dist-build/lang-model-tck-runner/src/test/java/org/jboss/weld/lang/model/tck/BuildtimeVetoExtension.java
OndroMih: None of these solutions would work because they would disable the `BuildCompatibleExtension` or one of its dependencies completely and the TC would fail. The solution in this PR is more of a hack but just works - the `BuildCompatibleExtension` would be loaded before the `ProcessAnnotatedType` in order to pass the TCK but then all its classes are vetoed so that they aren't processed later. If they are processed by WildFly later, an exception is thrown because some of the classes are package private and not available via reflection.
I think the only clean solutions would be either to veto each individual class in the CDI that is package private or make those classes public.

OndroMih:
Renamed wildfly8 to wildfly.

dist-build/lang-model-tck-runner/src/test/java/org/jboss/weld/lang/model/tck/BuildtimeVetoExtension.java
manovotn: It would work, the test relies on `LangModelVerifier` being discovered as a type (while not having any bean defining annotation). You can still set discovery to `annotated` and add the following method to the `BuildCompatibleExtension`:

```
    @Discovery
    public void addVerifier(ScannedClasses sc) {
        sc.add(LangModelVerifier.class.getName());
    }
```

With that, you should be able to remove `BuildtimeVetoExtension`. The reason I am trying to do that is because with it, the example isn't valid for any CDI Lite implementation trying to execute this TCK test. Lite impls do not have access to portable extension.

OndroMih: You're right, that worked. I've updated this PR along what you suggested.

PR #395, opened 2022-07-11
Revert "Disable the tck-audit profile by default so that the org.jbos…

### LINKED ISSUE #383: Version 4.0.5 fails on a missing dependency
It looks like `org.jboss.test-audit:jboss-test-audit-impl` is not available in the maven central:

```
Could not find org.jboss.test-audit:jboss-test-audit-impl:2.0.0.Final.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/org/jboss/test-audit/jboss-test-audit-impl/2.0.0.Final/jboss-test-audit-impl-2.0.0.Final.pom
```

It is available in the Jboss repo but it would be nice not to require the use of a custom repo.

manovotn:
…s.test-a… (#388)"

This reverts commit 22d9a76199356b8da8827ed76f415ef1ff45f107.

Fixes #383 

Puts us back in state where plain `mvn clean install` builds the whole project since we should no longer need jboss repo access to gain access to tck audit artifacts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,28 +3,9 @@
 
 Check out the [TCK Reference Guide](https://jakartaee.github.io/cdi-tck/) to get acquainted with the CDI TCK and learn how to execute and debug it.
 
-## Building CDI TCK artifacts
-To build the CDI TCK artifacts, use:
-
-`mvn -DtckAuditSourceGen install`
-
-or when compiling against staged Jakarta artifacts:
-
-`mvn -DtckAuditSourceGen -Pstaging install`
-
-## Building the CDI TCK distribution
-The CDI TCK distribution artifact is built by specifing an additional `-Drelease` property to build the TCK reference
-documentation and distribution bundle, e.g.:
-
-`mvn -DtckAuditSourceGen -Drelease install`
-
-## Eclipse Continuous Integration Environment
-The Eclipse continuous integration environment interface for the CDI project is located at https://ci.eclipse.org/cdi/
-The https://github.com/jakartaee/cdi/wiki/Eclipse-CI-Release-Jobs page describes the jobs found there.
-
 ## Sources in GIT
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -3,28 +3,9 @@
...
-`mvn -DtckAuditSourceGen -Drelease install`
-
-## Eclipse Continuous Integration Environment
-The Eclipse continuous integration environment interface for the CDI project is located at https://ci.eclipse.org/cdi/
-The https://github.com/jakartaee/cdi/wiki/Eclipse-CI-Release-Jobs page describes the jobs found there.
-
 ## Sources in GIT
 
-Master branch contains the CDI TCK 4.0
+Master branch contains the work-in-progress on CDI TCK 4.0
```
Ladicek, line 8: This isn't really true now, is it?

README.md
```diff
@@ -3,28 +3,9 @@
...
-## Building the CDI TCK distribution
-The CDI TCK distribution artifact is built by specifing an additional `-Drelease` property to build the TCK reference
-documentation and distribution bundle, e.g.:
-
-`mvn -DtckAuditSourceGen -Drelease install`
-
-## Eclipse Continuous Integration Environment
-The Eclipse continuous integration environment interface for the CDI project is located at https://ci.eclipse.org/cdi/
-The https://github.com/jakartaee/cdi/wiki/Eclipse-CI-Release-Jobs page describes the jobs found there.
-
```
Ladicek, line 24: Should this huge section really be removed? It seems to me it contains some useful info. I'd just remove `-DtckAuditSourceGen` from the commands and everything else would still make sense I think?

README.md
manovotn: @Ladicek good point, I really just did a revert of the original commit. I'll add these small bits back in.

Ladicek (approved):

Issue #389, opened 2022-06-24
TCK Challenge Tests with FailingHttpStatusCode exception


breakponchito:
**Specification:**
[cdi 4.0](https://jakarta.ee/specifications/cdi/4.0/)

**Challenged test(s):**
[org.jboss.cdi.tck.tests.context.application.async.ApplicationContextAsyncListenerTest#testApplicationContextActiveOnError](https://github.com/jakartaee/cdi-tck/blob/master/web/src/main/java/org/jboss/cdi/tck/tests/context/application/async/ApplicationContextAsyncListenerTest.java#L79)
[org.jboss.cdi.tck.tests.context.conversation.determination.ConversationDeterminationTest#testConversationDetermination](https://github.com/jakartaee/cdi-tck/blob/master/web/src/main/java/org/jboss/cdi/tck/tests/context/conversation/determination/ConversationDeterminationTest.java#L59)
[org.jboss.cdi.tck.tests.context.request.async.RequestContextAsyncListenerTest#testRequestContextActiveOnError](https://github.com/jakartaee/cdi-tck/blob/master/web/src/main/java/org/jboss/cdi/tck/tests/context/request/async/RequestContextAsyncListenerTest.java#L87)
[org.jboss.cdi.tck.tests.context.session.async.SessionContextAsyncListenerTest#testSessionContextActiveOnError](https://github.com/jakartaee/cdi-tck/blob/master/web/src/main/java/org/jboss/cdi/tck/tests/context/session/async/SessionContextAsyncListenerTest.java#L83)

**TCK version:**
4.0.5

**Tested implementation:**
[Payara 6 (branch)](https://github.com/breakponchito/Payara/tree/payara6-cdi-4.0-upgrade) with:

- WELD api: 5.0.Final
- WELD: 5.0.0.SP2
- CDI API: 4.0.1

**Description:**
The tests are failing because of the following exception thrown:  com.gargoylesoftware.htmlunit.FailingHttpStatusCodeException: 500 Internal Server Error

those tests are throwing ServletException as part of the flow to test and the Payara Server side is setting the 500 error, but on the TCK side with the htmlunit framework those errors are Wrapped by a new exception the FailingHttpStatusCodeException. This is the wrong behavior and we don't want to hack the server side to return different codes.

We saw a way to solve this but it is on the TCK side by setting the following property to false:  webClient.getOptions().setThrowExceptionOnFailingStatusCode(false)

With this value set to false all the failing tests will pass without hack the server side

[...]

PR #399, opened 2022-07-19
Fixed tests to disabling thrown of FailingHttpStatusCode exception

breakponchito:
The tests fixed here were designed to validate error page response. On some servers this is causing to thrown the FailingHttpStatusCode exception causing to fail the tests (Payara server is an example). This PR is to disable this behaviour that is part of the HtmlUnit implementation. This will help some servers to pass TCK by omitting the FailingHttpStatusCode exception

[challenge created](https://github.com/jakartaee/cdi-tck/issues/389)

### DIFF
--- a/web/src/main/java/org/jboss/cdi/tck/tests/context/application/async/ApplicationContextAsyncListenerTest.java
+++ b/web/src/main/java/org/jboss/cdi/tck/tests/context/application/async/ApplicationContextAsyncListenerTest.java
@@ -78,6 +78,7 @@ public class ApplicationContextAsyncListenerTest extends AbstractTest {
     @SpecAssertion(section = APPLICATION_CONTEXT_EE, id = "ae")
     public void testApplicationContextActiveOnError() throws Exception {
         WebClient webClient = new WebClient();
+        webClient.getOptions().setThrowExceptionOnFailingStatusCode(false);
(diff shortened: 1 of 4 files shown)


manovotn:
Fixes #389

manovotn (approved):

Ladicek:
LGTM, but I don't know what's the proper workflow when it comes to TCK challenges or so, so I'll leave that to you @manovotn :-)

manovotn:
I am not familiar with the flow from docs either, but I'd just go with making another release as we have this plus one other fix present already. Scott would know for sure but I think he's out of office now.

Also, as I stated in the issue, I don't think this is actually a valid challenge per se because the test doesn't care what's the HTTP status code returned. The test also exists for a long time in unchanged manner and I think what's causing these issues to emerge is a tooling upgrade done in https://github.com/jakartaee/cdi-tck/commit/0f9f0a484f10903e484b676e6bcea62ec97e91e3
Htmlunit now defaults to throwing an error where it was previously silently ignoring it.

So while not a challenge to what we assert, it is definitely a ground for improvement.

Ladicek:
Okay gotcha. This is a simple bugfix then. Makes sense.

breakponchito:
thank you guys to merge this

PR #401, opened 2022-07-27
Change AfterTypeDiscoveryTest to avoid assumptions about how many tot…

### LINKED ISSUE #400: Questions about why CDI TCK fails when Jakarta MVC / Eclipse Krazo are bundled in Glassfish 7
Hi all,

as the headline describes, I'd like to ask a few questions about specific CDI TCK tests which are failing when Jakarta MVC and Eclipse
Krazo are bundled in GF 7. I've done some analysis but I'm not sure if my conclusions are correct, so maybe someone here can help.

## ProcessBeanAttributesNotFiredForBuiltinBean

This test fails, because Krazo creates a `@RequestScoped` instance of `HttpServletRequest` in the [JaxRsContextProducer](https://github.com/eclipse-ee4j/krazo/blob/master/core/src/main/java/org/eclipse/krazo/jaxrs/JaxRsContextProducer.java) which is then observed by the TCK method `ProcessBeanAttributesObserver#observeHttpServletRequestBeanAttributes`. In this case I'm not sure if the producer in Krazo is really necessary / compliant with the CDI spec, because the container already provides a `@Default` scoped `HttpServletRequest`. 

Removing the producer method fixes the test, but I'd like to understand if this was some spec violation or maybe a missing scope / exclusion in the CDI TCK?

## AfterTypeDiscoveryTest

Here the methods `testFinalAlternatives` and `testInitialAlternatives` fail, because the `AfterTypeDiscoveryObserver` used as test extension contains a bunch of Krazo beans (screenshot) which are added in our [KrazoCdiExtension](https://github.com/eclipse-ee4j/krazo/blob/master/core/src/main/java/org/eclipse/krazo/cdi/KrazoCdiExtension.java). Most of them have in common, that the priority is set to a really low value, sometimes even `0`. In this case I'm not sure if the TCK needs to exclude classes which are not part of the test setup or if Krazo needs to prioritize its beans later.

Would be great if someone can help me and answer those questions. Even a hint what can be wrong would be helpful. Thanks in advance!

Screenshot from 2022-07-25 21-18-06

manovotn:
…al components are in test deployment.

Fixes #400

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/full/extensions/lifecycle/atd/AfterTypeDiscoveryTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/full/extensions/lifecycle/atd/AfterTypeDiscoveryTest.java
@@ -22,6 +22,7 @@ import static org.jboss.cdi.tck.cdi.Sections.PROCESS_PRODUCER;
 import static org.jboss.cdi.tck.cdi.Sections.TYPE_DISCOVERY_STEPS;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertTrue;
 
 import jakarta.enterprise.context.RequestScoped;
@@ -43,6 +44,8 @@ import org.jboss.test.audit.annotations.SpecAssertion;
 import org.jboss.test.audit.annotations.SpecAssertions;
 import org.jboss.test.audit.annotations.SpecVersion;
 import org.testng.annotations.Test;
+
+import java.util.List;
 
 /**
  * @author Martin Kouba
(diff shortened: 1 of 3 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/full/extensions/lifecycle/atd/AfterTypeDiscoveryTest.java
```diff
@@ -78,16 +81,35 @@ public void testInitialInterceptors() {
...
+        assertTrue(alphaAltIndex < echoAltIndex);
+        assertTrue(echoAltIndex < deltaAltIndex);
     }
 
     @Test
     @SpecAssertions({ @SpecAssertion(section = AFTER_TYPE_DISCOVERY, id = "d"), @SpecAssertion(section = AFTER_TYPE_DISCOVERY, id = "hc") })
     public void testInitialDecorators() {
-        assertEquals(extension.getDecorators().size(), 4);
+        assertTrue(extension.getDecorators().size() >= 4);
+        // TODO same issue?
```
manovotn, line 112: I'll perform similar change here for decorators as soon as we have confirmation that this was indeed the issue.

manovotn:
@erdlet This PR should fix your problems, do you think you could checkout this branch, build it (just `mvn clean install` will do) and test it with GF?

erdlet:
Thanks a lot for this fix. I'll test it until this evening against GF and provide feedback to you.

impl/src/main/java/org/jboss/cdi/tck/tests/full/extensions/lifecycle/atd/AfterTypeDiscoveryTest.java
Ladicek: Looking at `testInitialInterceptors` and it just asserts presence, do we really have to assert order for alternatives and decorators? (And if so, why not for interceptors? :-) )

manovotn: I think we check the ordering of interceptors in separate test (or rather, interceptors TCKs do that) but I haven't tried to pinpoint that, I just vaguely recall seeing such test.
It makes sense to assert ordering of alternatives/decorators as that is being manipulated in the extension. For instance decorator ordering is being reversed as part of the test.

I have altered the decorator logic in a similar way, so this PR should now be ready for review.
Also did a quick check with embedded Weld and it looks fine.

manovotn:
I have marked the PR and the issue as `challenge` and `accepted` for I believe in this case the test makes assumptions that it shouldn't. Nonetheless, if we verify that this fixes the problem for you, we can make a quick TCK release if we chose to :)

impl/src/main/java/org/jboss/cdi/tck/tests/full/extensions/lifecycle/atd/AfterTypeDiscoveryTest.java
Ladicek: OK, thanks. I was just wondering :-)

Ladicek (approved):

erdlet:
@manovotn I just ran the TCK built from your branch and ran it against GF 7 with MVC 2.1.0.RC1 and Krazo 3.0.0.RC1. 

It seems this test is broken too: `AfterTypeDiscoveryMassOperationsTest>Arquillian.run:138->testInitialAlternatives:77 expected [3] but found [11]`. 
I guess it takes the same assumptions as the test you already fixed. Can you please have a look at it too? The fixed test works like a charm :)

manovotn:
@erdlet good find!
Both tests should now be fixed as part of this PR :)

erdlet:
@manovotn thanks for the quick fix! I'll build and run the TCK again so I should've feedback in a few minutes if everything is fine now :)

erdlet:
@manovotn great, it works 👍🏻 Thanks a lot for your help!

> Nonetheless, if we verify that this fixes the problem for you, we can make a quick TCK release if we chose to

Would it be possible to release the TCK within the next days so I can create a PR to re-integrate MVC into GF? :)

manovotn:
@erdlet glad to hear that.
And yes, I'll do the release tomorrow and will let you know once its out.

PR #412, opened 2023-01-31
Move InjectableReferenceTest to full + move back to using decorators

Ladicek:
This test naturally belongs to the `full` group, because it verifies behavior of `BeanManager.getInjectableReference()`, which is only required to work in CDI Full. The `BeanContainer` interface does not define this method.

Furthermore, this test was previously modified to use an interceptor instead of a decorator. This commit restores the usage of a decorator to avoid confusing method name and specification reference.


Ladicek:
The PR in which the test was changed previously is #292.

I also tried to run this with Weld and it passes just fine.

manovotn (approved):
I no longer recall what was the idea behind using interceptors but it surely belongs to Full, so +1

Issue #421, opened 2023-02-10
org.jboss.cdi.tck.tests.interceptors.definition.InterceptorDefinitionTest


jeanouii:
Per the current [TCK Process documentation](https://jakarta.ee/committees/specification/tckprocess/), please find bellow a challenge related to CDI TCK. Can someone please add the `challenge` label which is required by the TCK Process but not available to everyone to add.

### The relevant specification version and section number(s)

https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html
Section 1.2.5. Relationship to Jakarta Interceptors

https://jakarta.ee/specifications/interceptors/2.1/jakarta-interceptors-spec-2.1.html
Section 5.2.1. Use of the Priority Annotation in Ordering Interceptors

### The coordinates of the challenged test(s)

Only following test in the class is related to the challenge 

* testResolveInterceptorsReturnsOrderedList

### The exact TCK version

CDI TCK version 4.0.7

### The implementation being tested, including name and company

Apache OpenWebBeans 4.0.0-SNAPSHOT from the Apache Software Foundation

### A full description of why the test is invalid and what the correct behavior is believed to be

The test deploys a web application with 3 interceptors with a @Priory(APPLICATION). Then it does use the API to resolve interceptors based on a type and a set of annotations (bindings).

The specification says

> 2.9.1.10. Interceptor resolution
> The method BeanContainer.resolveInterceptors() returns the ordered list of interceptors for a set of interceptor bindings and a type of interception and which are enabled in the module or library containing the class into which the BeanContainer was injected, as defined in [Interceptor resolution](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#interceptor_resolution).

It says "ordered list" with giving much information what's expected in terms of ordering

1. the specification should be more specific on the expected order like it does somewhere else
2. the test should not assume a specific order

To me for consistency reason, we should use the @Priority rules to order the list. 

### Any supporting material; debug logs, test output, test logs, run scripts, etc.

> [ERROR]   InterceptorDefinitionTest>Arquillian.run:138->testResolveInterceptorsReturnsOrderedList:159 expected [true] but found [false]

Issue #420, opened 2023-02-10
org.jboss.cdi.tck.interceptors.tests.bindings.overriding.InterceptorBindingOverridingTest


jeanouii:
Per the current [TCK Process documentation](https://jakarta.ee/committees/specification/tckprocess/), please find bellow a challenge related to CDI TCK. Can someone please add the `challenge` label which is required by the TCK Process but not available to everyone to add.

### The relevant specification version and section number(s)

https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html
Section 1.2.5. Relationship to Jakarta Interceptors

https://jakarta.ee/specifications/interceptors/2.1/jakarta-interceptors-spec-2.1.html
Section 5.2.1. Use of the Priority Annotation in Ordering Interceptors

### The coordinates of the challenged test(s)

Only following test in the class is related to the challenge 

* testInterceptorBindingOverriden

### The exact TCK version

CDI TCK version 4.0.7

### The implementation being tested, including name and company

Apache OpenWebBeans 4.0.0-SNAPSHOT from the Apache Software Foundation

### A full description of why the test is invalid and what the correct behavior is believed to be

The test is composed of a simple bean Pony with 2 classes level interceptors and 1 method level interceptor overriding one of the 2 class lever interceptor. So in the end, we have a list of 2 interceptors, one being the method interceptor and the other one is the remaining class level interceptor (because the other one got overriden). 

Section 5.2. Interceptor Ordering Rules of Interceptor 2.1 also refers to ordering using @Priority annotation and says

> All interceptors specified using interceptor binding annotations visible on the target class (e.g., specified on the class or visible on the class because they were declared with the Inherited annotation) are combined with all binding annotations on the target method and sorted by the priorities specified by the Priority annotation; and then the interceptor methods are invoked in order of priority

In this sentence the last part explicitely says the final list needs to be sorted by the @Priority annotation. And once again, all same priority so not deterministic order.

````
        // getAge() returns 3, SlowAgingInterceptor adds one (4) and NegatingInterceptor negates (-4)
        assertEquals(pony.getAge(), -4);
````

This could be first negate 3 so you get -3 and add 1 so you get -2 and not -4.

### Any supporting material; debug logs, test output, test logs, run scripts, etc.

> [ERROR]   InterceptorBindingOverridingTest>Arquillian.run:138->testInterceptorBindingOverriden:61 expected [-4] but found [-2]

mkouba:
[...]

Issue #419, opened 2023-02-10
org.jboss.cdi.tck.interceptors.tests.contract.lifecycleCallback.bindings.LifecycleInterceptorDefinitionTest


jeanouii:
Per the current [TCK Process documentation](https://jakarta.ee/committees/specification/tckprocess/), please find bellow a challenge related to CDI TCK. Can someone please add the `challenge` label which is required by the TCK Process but not available to everyone to add.

### The relevant specification version and section number(s)

https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html
Section 1.2.5. Relationship to Jakarta Interceptors

https://jakarta.ee/specifications/interceptors/2.1/jakarta-interceptors-spec-2.1.html
Section 5.2.1. Use of the Priority Annotation in Ordering Interceptors

### The coordinates of the challenged test(s)

Only following test in the class is related to the challenge 

* testMultipleLifecycleInterceptors

### The exact TCK version

CDI TCK version 4.0.7

### The implementation being tested, including name and company

Apache OpenWebBeans 4.0.0-SNAPSHOT from the Apache Software Foundation

### A full description of why the test is invalid and what the correct behavior is believed to be

In the test mentioned above, at least 2 interceptors in the stack have `@Priority(APPLICATION)`, `AirborneInterceptor,` and `DestructionInterceptor,`.

Per the interceptor specification 2.1, Section 5.2.1, it's explicitely written `If more than one interceptor has the same priority, the relative order of those interceptors is undefined.`. The 2 tests mentioned above assert a specific order which is against the specification.

Section 5.2. Interceptor Ordering Rules of Interceptor 2.1 also refers to ordering using @Priority annotation. 

### Any supporting material; debug logs, test output, test logs, run scripts, etc.

> [ERROR]   LifecycleInterceptorDefinitionTest>Arquillian.run:138->testMultipleLifecycleInterceptors:83 ActionSequence [name=postConstruct, data=[SuperDestructionInterceptor, DestructionInterceptor, AirborneInterceptor, Weapon, Rocket]] and expected sequence differ on the index 0. expected [AirborneInterceptor] but found [SuperDestructionInterceptor]

Issue #418, opened 2023-02-10
org.jboss.cdi.tck.interceptors.tests.bindings.aroundConstruct.ConstructorInterceptionTest


jeanouii:
Per the current [TCK Process documentation](https://jakarta.ee/committees/specification/tckprocess/), please find bellow a challenge related to CDI TCK. Can someone please add the `challenge` label which is required by the TCK Process but not available to everyone to add.

### The relevant specification version and section number(s)

https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html
Section 1.2.5. Relationship to Jakarta Interceptors

https://jakarta.ee/specifications/interceptors/2.1/jakarta-interceptors-spec-2.1.html
Section 5.2.1. Use of the Priority Annotation in Ordering Interceptors

### The coordinates of the challenged test(s)

Only following tests in the class are related to the challenge 

* testMultipleConstructorLevelBinding
* testTypeLevelAndConstructorLevelBinding

### The exact TCK version

CDI TCK version 4.0.7

### The implementation being tested, including name and company

Apache OpenWebBeans 4.0.0-SNAPSHOT from the Apache Software Foundation

### A full description of why the test is invalid and what the correct behavior is believed to be

In both of the tests mentioned above, at least 2 interceptors in the stack have `@Priority(2000)`

Per the interceptor specification 2.1, Section 5.2.1, it's explicitely written `If more than one interceptor has the same priority, the relative order of those interceptors is undefined.`. The 2 tests mentioned above assert a specific order which is against the specification.

NOTE: in CDI some sections like, 2.8.5.2 which relates to Observer ordering also have a similar mentioned, `The order of more than one observer with the same priority is undefined and the observer methods are notified therefore in a non predictable order.` which is consistent.

### Any supporting material; debug logs, test output, test logs, run scripts, etc.

> [ERROR]   ConstructorInterceptionTest>Arquillian.run:138->testMultipleConstructorLevelBinding:72 ActionSequence [name=default, data=[BravoInterceptor, AlphaInterceptor2, BeanWithMultipleConstructorLevelBinding]] and expected sequence differ on the index 0. expected [AlphaInterceptor2] but found [BravoInterceptor]
[...]

PR #422, opened 2023-02-13
Issue #418 #419 #420 #421 Deterministic interceptor order (@Priority)

jeanouii:
This small PR makes the interceptor order deterministic because the asserts as made such as order is implied.


manovotn (approved):
Looks good.
I've also verified this with Weld/WFLY.

jeanouii:
Wow, thanks a lot

Issue #424, opened 2023-02-13
org.jboss.cdi.tck.tests.full.context.passivating.dependency.builtin.BuiltinBeanPassivationDependencyTest


jeanouii:
Per the current [TCK Process documentation](https://jakarta.ee/committees/specification/tckprocess/), please find bellow a challenge related to CDI TCK. Can someone please add the `challenge` label which is required by the TCK Process but not available to everyone to add.

### The relevant specification version and section number(s)

https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html

### The coordinates of the challenged test(s)

Only following test in the class are related to the challenge 

* testInjectionPoint

### The exact TCK version

CDI TCK version 4.0.7

### The implementation being tested, including name and company

Apache OpenWebBeans 4.0.0-SNAPSHOT from the Apache Software Foundation

### A full description of why the test is invalid and what the correct behavior is believed to be

The test essentially passivates (serialize) a bean instance and activates it back (deserialize). Then it asserts that the bean is the same. When comparing the bean instance before serialization and the bean instance after deserialization, the test assumes Annotated implements equals() which isn't required or specified anywhere in the spec or in the javadoc.

> assertEquals(inspectorCopy.getInjectionPoint().getAnnotated(), inspector.getInjectionPoint().getAnnotated());

The section 3.9.4. Alternative metadata sources defines the interface but does not require equals()/hashcode() to be implemented. It only says 

> The interface jakarta.enterprise.inject.spi.Annotated exposes the overriding annotations and type declarations.

````
public interface Annotated {
    Type getBaseType();
    Set<Type> getTypeClosure();
    <T extends Annotation> T getAnnotation(Class<T> annotationType);
    <T extends Annotation> Set<T> getAnnotations(Class<T> annotationType);
    Set<Annotation> getAnnotations();
    boolean isAnnotationPresent(Class<? extends Annotation> annotationType);
}
````

### Any supporting material; debug logs, test output, test logs, run scripts, etc.

> java.lang.AssertionError:
Expected :Annotated Field 'inspector', Base Type : class org.jboss.cdi.tck.tests.full.context.passivating.dependency.builtin.Inspector,Type Closures : null,Annotations : [@jakarta.inject.Inject()],Java Member Name : inspector
[...]

PR #430, opened 2023-02-20
Issue #424 passivation/activation and Annotated equality

jeanouii:
This small PR addresses the challenge of #424

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/full/context/passivating/dependency/builtin/BuiltinBeanPassivationDependencyTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/full/context/passivating/dependency/builtin/BuiltinBeanPassivationDependencyTest.java
@@ -97,7 +97,11 @@ public class BuiltinBeanPassivationDependencyTest extends AbstractTest {
         assertEquals(inspectorCopy.getInjectionPoint().getQualifiers(), inspector.getInjectionPoint().getQualifiers());
         assertEquals(inspectorCopy.getInjectionPoint().getBean(), inspector.getInjectionPoint().getBean());
         assertEquals(inspectorCopy.getInjectionPoint().getMember(), inspector.getInjectionPoint().getMember());
-        assertEquals(inspectorCopy.getInjectionPoint().getAnnotated(), inspector.getInjectionPoint().getAnnotated());
+
+        // Annotated does not necessarly implement equals()/hashcode()
+        assertEquals(inspectorCopy.getInjectionPoint().getAnnotated().getBaseType(), inspector.getInjectionPoint().getAnnotated().getBaseType());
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/full/context/passivating/dependency/builtin/BuiltinBeanPassivationDependencyTest.java
```diff
@@ -97,7 +97,11 @@ public void testInjectionPoint() throws IOException, ClassNotFoundException {
         assertEquals(inspectorCopy.getInjectionPoint().getQualifiers(), inspector.getInjectionPoint().getQualifiers());
         assertEquals(inspectorCopy.getInjectionPoint().getBean(), inspector.getInjectionPoint().getBean());
         assertEquals(inspectorCopy.getInjectionPoint().getMember(), inspector.getInjectionPoint().getMember());
-        assertEquals(inspectorCopy.getInjectionPoint().getAnnotated(), inspector.getInjectionPoint().getAnnotated());
+
+        // Annotated does not necessarly implement equals()/hashcode()
+        assertEquals(inspectorCopy.getInjectionPoint().getAnnotated().getBaseType(), inspector.getInjectionPoint().getAnnotated().getBaseType());
+        assertEquals(inspectorCopy.getInjectionPoint().getAnnotated().getAnnotations(), inspector.getInjectionPoint().getAnnotated().getAnnotations());
```
Ladicek, line 103: These are both good, but I think we should also (mainly?) assert that the underlying reflection objects are equal (which should work?). I know we already assert equality on `InjectionPoint.getMember()`, but I do think we should have both.

Something like

```java
assertEquals(
    unwrap(inspectorCopy.getInjectionPoint().getAnnotated()),
    unwrap(inspector.getInjectionPoint().getAnnotated())
);
```

where

```java
private static Object unwrap(Annotated annotated) {
    if (annotated instanceof AnnotatedMember) {
        return ((AnnotatedMember) annotated).getJavaMember();
    } else if (annotated instanceof AnnotatedParameter) {
        return ((AnnotatedParameter) annotated).getJavaParameter();
    } else if (annotated instanceof AnnotatedType) {
        return ((AnnotatedType) annotated).getJavaClass();
    } else {
        throw new UnsupportedOperationException("Unknown Annotated instance: " + annotated);
    }
}
```

WDYT?

jeanouii: Oh I see.
I'm happy to update the PR according to this. It looks more complete.

jeanouii:
Weird, but yes I'll see if I can get ride of the last commit

jeanouii:
Hopefully it's better. I extracted the method into the assert util class so it can be reused somewhere else if needed as your proposal was a bit more generic than what we needed

jeanouii:
Added them back for completeness.

Ladicek (approved):

manovotn (approved):
Looks good, thanks!

Issue #429, opened 2023-02-14
org.jboss.cdi.tck.tests.definition.bean.BeanDefinitionTest


jeanouii:
Per the current [TCK Process documentation](https://jakarta.ee/committees/specification/tckprocess/), please find bellow a challenge related to CDI TCK. Can someone please add the `challenge` label which is required by the TCK Process but not available to everyone to add.

### The relevant specification version and section number(s)

https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#bean_types
https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#managed_bean_types

### The coordinates of the challenged test(s)

Only following test in the class are related to the challenge 

* testRawBeanTypes

### The exact TCK version

CDI TCK version 4.0.7

### The implementation being tested, including name and company

Apache OpenWebBeans 4.0.0-SNAPSHOT from the Apache Software Foundation

### A full description of why the test is invalid and what the correct behavior is believed to be

The specification defines the following 

> A bean type defines a client-visible type of the bean. A bean may have multiple bean types.

And then provides a list of valid and invalid type. But I'm facing a case where it's ambiguous, at least I can find different behaviors so it's worth either updating a bit the TCK or improving the specification.

The test may fail with the assert on MySuperInterface.

````
    @Test
    @SpecAssertions({ @SpecAssertion(section = MANAGED_BEAN_TYPES, id = "a"), @SpecAssertion(section = BEAN_TYPES, id = "a"),
            @SpecAssertion(section = BEAN_TYPES, id = "l"), @SpecAssertion(section = BEAN, id = "ba") })
    public void testRawBeanTypes() {
        assertEquals(getBeans(MyRawBean.class).size(), 1);
        Bean<MyRawBean> bean = getBeans(MyRawBean.class).iterator().next();
        assertEquals(bean.getTypes().size(), 5);
        assertTrue(bean.getTypes().contains(MyRawBean.class));
        assertTrue(bean.getTypes().contains(MyBean.class));
        assertTrue(bean.getTypes().contains(MyInterface.class));
        assertTrue(bean.getTypes().contains(MySuperInterface.class));
        assertTrue(bean.getTypes().contains(Object.class));
    }
````

````
interface MyInterface extends MySuperInterface<Number> {
}
````

[...]

PR #433, opened 2023-02-24
Issue #429 Tolerate both MySuperInterface

jeanouii:
and MySuperInterface<Number> for raw-type types.

Tested with Weld and OpenWebBeans

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/definition/bean/BeanDefinitionTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/definition/bean/BeanDefinitionTest.java
@@ -130,7 +130,12 @@ public class BeanDefinitionTest extends AbstractTest {
         assertTrue(bean.getTypes().contains(MyRawBean.class));
         assertTrue(bean.getTypes().contains(MyBean.class));
         assertTrue(bean.getTypes().contains(MyInterface.class));
-        assertTrue(bean.getTypes().contains(MySuperInterface.class));
+
+        // CDI isn't clear whereas we should get MySuperInterface with or without Number parameter.
+        // It assumes JLS must be followed to determine the list of super types
+        // JLS isn't clear itself and Java implementations either. So we need to keep it open and support both
+        // https://bugs.openjdk.org/browse/JDK-8044366
+        assertTrue(containsClass(bean.getTypes(), MySuperInterface.class));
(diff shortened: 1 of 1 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/definition/bean/BeanDefinitionTest.java
```diff
@@ -130,7 +130,12 @@ public void testRawBeanTypes() {
         assertTrue(bean.getTypes().contains(MyRawBean.class));
         assertTrue(bean.getTypes().contains(MyBean.class));
         assertTrue(bean.getTypes().contains(MyInterface.class));
-        assertTrue(bean.getTypes().contains(MySuperInterface.class));
+
+        // CDI isn't clear whereas we should get MySuperInterface with or without Number parameter.
+        // It assumes JLS must be followed to determine the list of super types
+        // JLS isn't clear itself and Java implementations either. So we need to keep it open and support both
+        // https://bugs.openjdk.org/browse/JDK-8044366
+        assertTrue(containsClass(bean.getTypes(), MySuperInterface.class));
```
Ladicek, line 138: I understand the `containsClass` helper method is used as a shortcut, and it's used like that elsewhere, so this is probably fine.

I'd just like to point out that, per my understanding of the JLS ambiguity, `MyRawBean` may have either raw `MySuperInterface` or parameterized `MySuperInterface<Number>` as a superinterface type. This test would consider `MySuperInterface<AnythingElse>` correct too. Myself, I'd do something like

```java
assertTrue(bean.getTypes().contains(MySuperInterface.class)
        || bean.getTypes().contains(new TypeLiteral<MySuperInterface<Number>>() {}.getType()));
```

But as mentioned above, the `containsClass` shortcut is used elsewhere too, and it's unlikely CDI implementations would construct `MySuperInterface<SomethingElse>` in this case, so I guess we can keep it as is.

manovotn: > This test would consider MySuperInterface<AnythingElse> correct too. Myself, I'd do something like

That's a good point. For the sake of clarity we should rather use what Ladislav suggests to make it explicit what two types are accepted.

Ladicek (approved):

PR #436, opened 2023-02-27
Cleanup excess of code from old versions

jeanouii:
As discussed in the mailing list, here is a very small PR to remove the excess of code due to old versions of the TCK


manovotn:
In short, this code was a dead code (apart from observing if the event happens) since addition due to a copy-paste from Weld test suite.
Detailed explanation is in this email - https://www.eclipse.org/lists/cdi-dev/msg00726.html

manovotn (approved):

PR #437, opened 2023-03-01
Add Support for Selenium Tests

brideck:
- Adds Selenium implementation for Faces Ajax test

### DIFF
--- a/doc/reference/src/main/asciidoc/executing.asciidoc
+++ b/doc/reference/src/main/asciidoc/executing.asciidoc
@@ -83,6 +83,13 @@ The Arquillian will also start and stop the application server automatically (pr
 Since Arquillian in-container tests are executed in a remote JVM, the results of the test must be communicated back to the runner over a container-supported protocol. The TCK utilizes servlet-based protocol (communication over HTTP). 
 
 
+[NOTE]
+====
+Some implementations of Jakarta Faces use modern Javascript that is incompatible with the TCK's default test tooling. In order to test these implementations, specify the property -Drun.selenium=true. An installation of Chrome is required.
+
+====
+
+
 === Running the Tests In the Container - SE part
 
 To execute full TCK testsuite you have to run tests within SE group as well. 
--- a/pom.xml
+++ b/pom.xml
@@ -112,6 +112,7 @@
         <arquillian.container.se.api.version>1.0.1.Final</arquillian.container.se.api.version>
         <apache.httpclient.version>3.1</apache.httpclient.version>
         <htmlunit.version>2.50.0</htmlunit.version>
+	<selenium.version>4.7.2</selenium.version>
         <!-- Required for distribution build, should be overriden for each build -->
         <cdi.tck.version>${project.version}</cdi.tck.version>
         <weld.version>5.0.0.SP2</weld.version>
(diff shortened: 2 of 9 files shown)


Emily-Jiang:
@manovotn @Ladicek can you take a look at the PR?

Emily-Jiang (approved):
LGTM

Emily-Jiang:
Any other review comments @manovotn @Ladicek @starksm64 ? Appreciate your feedback. Thanks in advance!

pnicolucci:
@Emily-Jiang some licensing discussion: https://github.com/jakartaee/faces/pull/1794#issuecomment-1453392777

pnicolucci:
@Emily-Jiang @brideck license discussion has finalized: https://github.com/jakartaee/faces/pull/1794 contains the details. This PR is good to go with the Apache copyright header.

manovotn (commented):
> I plan to merge this PR tomorrow morning. If you have any objections, please shout asap.

Shouldn't we instead look to remove htmlunit altogether and replace it with selenium tests?
This PR only adds a special-case execution for a singular test.

What are the requirements on impl side to execute such test?
In fact, I think there should be some documentation in terms of requirements for TCK execution and what setup is needed in the [TCK doc](https://jakartaee.github.io/cdi-tck/).

> @starksm64 just a heads up, I will need some help from you to do a service release after this PR is merged.

@Emily-Jiang as a committer you are able to access the Jenkins yourself and run the TCK release job followed up by the maven repo release job. I assume it's even documented somewhere but I have no idea where from the top of my head :)

brideck:
> 
> Shouldn't we instead look to remove htmlunit altogether and replace it with selenium tests? This PR only adds a special-case execution for a singular test.

I don't think that a wholesale technology replacement would be an appropriate change for a TCK service release. The change is designed to provide zero impact to implementations that are already passing the TCK as-is. They would just continue to run with no changes, without adding the new property.

> 
> What are the requirements on impl side to execute such test? In fact, I think there should be some documentation in terms of requirements for TCK execution and what setup is needed in the [TCK doc](https://jakartaee.github.io/cdi-tck/).
 
We could certainly add something to the documentation. Because the logic is the same between the two test implementations, the idea is that someone would only need run one or the other, depending on the modernity of their Faces implementation's javascript support. The changes are that someone running the Selenium tests would need to set the property in their TCK runner and also provide Chrome in their test environment.

brideck:
I've added a note to the documentation explaining why and how you would want to use this new support.

manovotn:
> I don't think that a wholesale technology replacement would be an appropriate change for a TCK service release. The change is designed to provide zero impact to implementations that are already passing the TCK as-is. They would just continue to run with no changes, without adding the new property.

Fair enough, I've created https://github.com/jakartaee/cdi-tck/issues/441 to track the future change.

> I've added a note to the documentation explaining why and how you would want to use this new support.

Thanks!

manovotn (approved):
I don't fully understand the internals of the problem but the solution is well isolated and doesn't affects other impls testing it, hence my approval :)

Emily-Jiang:
Thank you @manovotn @Ladicek for reviewing this! I will try to follow the instruction and perform a service release shortly.

Issue #431, opened 2023-02-21
org.jboss.cdi.tck.tests.full.extensions.lifecycle.bbd.broken.passivatingScope.AddingPassivatingScopeTest


jeanouii:
Per the current [TCK Process documentation](https://jakarta.ee/committees/specification/tckprocess/), please find bellow a challenge related to CDI TCK. Can someone please add the `challenge` label which is required by the TCK Process but not available to everyone to add.

### The relevant specification version and section number(s)

https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#passivation_validation
https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#exceptions

### The coordinates of the challenged test(s)

Only following test in the class are related to the challenge 

* testAddingScopeType

### The exact TCK version

CDI TCK version 4.0.7

### The implementation being tested, including name and company

Apache OpenWebBeans 4.0.0-SNAPSHOT from the Apache Software Foundation

### A full description of why the test is invalid and what the correct behavior is believed to be

This test add a passivating scope to a non serializable bean. Per validation rules defined in https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#passivation_validation this is not legal and must result in a deployment error.

>  If a managed bean which declares a passivating scope, or a built-in bean:
>
> * is not passivation capable,
> * has an injection point that is not passivation capable,
> * has an interceptor or decorator that is not passivation capable,
> * has an interceptor or decorator with an injection point that is not passivation capable,
>
> then the container automatically detects the problem and treats it as a deployment problem.

From https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#exceptions

> Deployment problems - occur when there are problems resolving dependencies, or inconsistent specialization (in CDI Full), in a particular deployment. If a deployment problem occurs, the container must throw a subclass of jakarta.enterprise.inject.spi.DeploymentException.

The test is asserting the wrong exception jakarta.enterprise.inject.spi.DefinitionException

### Any supporting material; debug logs, test output, test logs, run scripts, etc.

````
[...]

PR #439, opened 2023-03-02
Issue #438 Fix the interceptor ordering test for both CDI Lite and Full

### LINKED ISSUE #438: org.jboss.cdi.tck.tests.interceptors.definition.interceptorOrder.InterceptorOrderTest
Due to the `cdi-full` TCK separation, this test is now invalid:

1. `testInterceptorsCalledInOrderDefinedByBeansXml` actually verifies `@Priority` ordering,
2. `testInterceptorsInvocationOrder` uses `@AroundInvoke` interceptors declared on target beans, which doesn't have to be supported by CDI Lite implementations.

I propose to:

1. Create a copy in the `full` package (with the `cdi-full` category) that restores the original test, including the `beans.xml` file.
2. In the current test, remove the `testInterceptorsInvocationOrder` method and then rename `testInterceptorsCalledInOrderDefinedByBeansXml` to `testInterceptorsInvocationOrder`.

Ladicek:
Fixes #438

### DIFF
--- a/impl/src/main/resources/tck-audit-cdi.xml
+++ b/impl/src/main/resources/tck-audit-cdi.xml
@@ -5157,6 +5157,13 @@
             </text>
         </assertion>
 
+        <assertion id="l">
+            <text>The Priority annotation can be used to enable and order interceptors associated with components that use interceptor bindings.
+                The required value element of the Priority annotation determines the ordering.
+                Interceptors with smaller priority values are called first.</text>
+            <note>This is defined in the Interceptors 2.0 spec, 5.2.1. Use of the Priority Annotation in Ordering Interceptors.</note>
+        </assertion>
+
     </section>
 
     <section id="interceptor_resolution" title="Interceptor resolution" level="2">
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/definition/interceptorOrder/AccountBinding.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/definition/interceptorOrder/AccountBinding.java
@@ -1,4 +1,4 @@
-package org.jboss.cdi.tck.tests.interceptors.definition.interceptorOrder;
+package org.jboss.cdi.tck.tests.full.interceptors.definition.interceptorOrder;
 
 import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/definition/interceptorOrder/AccountTransaction.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/definition/interceptorOrder/AccountTransaction.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source
- * Copyright 2012, Red Hat, Inc., and individual contributors
+ * Copyright 2010, Red Hat, Inc., and individual contributors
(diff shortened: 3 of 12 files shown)


impl/src/main/resources/tck-audit-cdi.xml
```diff
@@ -5157,6 +5157,13 @@
             </text>
         </assertion>
 
+        <assertion id="l">
+            <text>The Priority annotation can be used to enable and order interceptors associated with components that use interceptor bindings.
```
manovotn, line 5161: I don't mind this, just FTR - there is a `tck-audit-int` file which will probably already have a section related to this.

manovotn (approved):

impl/src/main/resources/tck-audit-cdi.xml
Ladicek: `tck-audit-int` has an `id-`less assertion in `priority_annotation_in_ordering_int` section, with a note that "Tested within the CDI TCK, section 9.4 Interceptor enablement and ordering", so I'm not entirely sure: should I quote this assertion from `tck-audit-int` instead? Adding a new assertion to the "Interceptor enablement and ordering" section of `tck-audit-cdi` seemed a bit better to me, but I'm not really sure.

manovotn: I see In that case, the current solution seems good to me

jeanouii:
Looks good

Emily-Jiang:
In light of recent TCK runs, it was evident that this fix has caused a previous compatible implementation no longer passes this test. According to the [Jakarta TCK process](https://jakarta.ee/committees/specification/tckprocess/), the accepted changes normally just exclude tests or expanding tests but not completely changing tests. 

> **Accepted Challenges**
> 
> A consensus that a test produces invalid results will result in the exclusion of that test from certification requirements, and an immediate update and release of an official distribution of the TCK including the new exclude list. The associated challenge issue MUST be closed with an accepted label to indicate it has been resolved.
> 
> The specification project may approve (user) workarounds for an accepted TCK challenge (as an alternative to excluding TCK tests).
> 
> As another alternative to excluding a challenged test, it may be possible to adjust the test validation logic to “expand” the validation check. E.g. if a test expects a value “A1” for a specific variable “x”, but a challenge is raised arguing that the specification language actually allows for either values “A1” and “A2” (but no other values) to be valid values for “x”, then it could be a valid course of action to modify the challenged test to allow either “A1” OR “A2” for “x”.
> 
> Since this line of thinking might be applied to cases that aren’t quite as straightforward as this example, care should be taken when using this approach. A particular danger is that an implementation that has already demonstrated compliance before the challenge was raised might actually not pass the new, modified test.
> 
> To limit the confusion and additional work such a scenario would cause, if there is already at least one certified compatible implementation before the challenge, the new, modified TCK should be run against at least one such implementation (and ideally all of them) before the changes are published, released, and finalized.
> 
> If such a change were released, and it was later found to cause a previously-certified implementation to fail the new, modified test, then excluding the test would likely be the only option, and this would require yet another, additional service release.
If this PR will be released under a major or minor release, it is fine. However, it does not comply with the TCK process if it is made to a service release.

Ladicek:
I believe your comment belongs to #431 and not here. Correct?

Emily-Jiang:
> I believe your comment belongs to #431 and not here. Correct?

oops. you were right. sorry about this.

Ladicek:
No worries, I was just wondering, because I'm pretty sure I verified this PR with Weld prior to submitting.

PR #443, opened 2023-03-07
Fix invalid annotation literals

Ladicek:
The `Annotation.equals()` contract directly states that for two annotation objects to be equal, they must (among others) both implement the [same] annotation interface.

The CDI TCK used to create instances of memberless annotations simply by instantiating an anonymous subclass of `AnnotationLiteral`, which doesn't implement the annotation interface and hence is invalid.

It has never been a problem, because `AnnotationLiteral` implements equality in a slightly more permissive manner: instead of checking that both objects implement the same annotation interface, it checks that both objects are annotations that have an equal `annotationType()`. It becomes problematic when an implementation uses a different mechanism to instantiate annotations than extending `AnnotationLiteral`, one that closely follows the `Annotation` contract. 

This commit replaces all usages of such anonymous subclasses by proper `AnnotationLiteral` subclasses that implement the annotation interface as well.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/bindings/resolution/ConstructorBinding.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/bindings/resolution/ConstructorBinding.java
@@ -25,6 +25,7 @@ import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
+import jakarta.enterprise.util.AnnotationLiteral;
 import jakarta.interceptor.InterceptorBinding;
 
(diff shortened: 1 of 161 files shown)


Ladicek:
Related to https://github.com/jakartaee/cdi/issues/656

I'm not submitting this as a challenge, because we decided to compensate for this on our side, but I believe this should be added to the next major/minor CDI TCK version (4.1/5.0 I believe).

manovotn (approved):
Thanks for looking into this!

manovotn:
Moving into ready for review state as 4.1 changes can go into main branch now.

Ladicek:
This has been approved before, so I'll just push the button :-)

PR #444, opened 2023-03-07
431: Disable Breaking Test

brideck:
The changes made for https://github.com/jakartaee/cdi-tck/issues/431 do not work with Weld, which has already passed the TCK on a previous release. Per the Jakarta EE TCK Process, this type of change is not allowed in response to an accepted challenge.

Disabling the test instead, so all implementations can pass the TCK.

### DIFF
--- a/impl/src/main/resources/tck-tests.xml
+++ b/impl/src/main/resources/tck-tests.xml
@@ -23,7 +23,12 @@
         </packages>
 
         <classes>
-            <!-- Add excluded tests here -->
+            <!-- CDITCK-431 -->
+            <class name="org.jboss.cdi.tck.tests.full.extensions.lifecycle.bbd.broken.passivatingScope.AddingPassivatingScopeTest">
+                <methods>
+                    <exclude name=".*"/>
+                </methods>
+            </class>
         </classes>
 
     </test>


manovotn:
> Per the [Jakarta EE TCK Process](https://jakarta.ee/committees/specification/tckprocess/), the fix for this challenge is inappropriate as it (knowingly) breaks an implementation (Weld) that has already passed the TCK.

Technically, this is correct because Weld has wrong implementation - in this case returns wrong exception.
I have an issue for it - https://issues.redhat.com/browse/WELD-2741 with trivial fix but it effectively needs a TCK release so that I can bump TCK version in Weld.

I have no problem releasing new version of Weld soon so we can technically re-enable this test as soon as we release the next TCK version to Central, right?

Emily-Jiang:
@manovotn What you suggested makes sense. We exclude this test for now so that we can release 4.0.8 and then reenable this test for the next service release once you has a fix for Weld.

Emily-Jiang (approved):
LGTM

manovotn:
@brideck @Emily-Jiang please send a PR re-enabling it once you're done with TCK release.

Emily-Jiang:
> @brideck @Emily-Jiang please send a PR re-enabling it once you're done with TCK release.

It is okay to reenable this test on the master branch but this fix should not be made to a service release. Please see more explaination [here](https://github.com/jakartaee/cdi-tck/pull/439#issuecomment-1460404507).

manovotn:
@Emily-Jiang not sure I understand - the parts you quoted do not say the exact same version of implementation must satisfy all minor releases. So current Weld version will pass TCK 4.0.8 at which point I can release Weld 5.1.1 that will pass this test so if you enable it, you are back at passing existing impl?

brideck:
> This is a rather formalistic approach. Clearly Weld has a bug and doesn't satisfy the specification requirements. Coincidentally, the TCK had the same bug. In the end, both must be fixed. Personally, I don't mind implementations excluding this TCK test (it was buggy after all), but I also don't mind fixing it in a service release.

The problem with trying to fix it in a service release comes with synchronizing those two changes, and making sure that all EE platform implementers are even aware of them. In this case, since the TCK was updated first, Open Liberty was broken with no possible means to get the test working again, as there is no version of Weld containing the necessary matching fix. Any other platform implementation using Weld for CDI would be in the same painful boat. Only making changes like this at the minor/major release barrier ensures you're not willfully breaking those folks.

Ladicek:
Well, _someone_ has to exclude the test. We have [at least] 2 implementations, one that is buggy but passes a buggy TCK test, and other that is correct but fails due to a buggy TCK test. Requiring the correct implementation to exclude because an incorrect implementation got to pass the TCK first is hardly fair.

My suggestion would be, let's just exclude the test for good and reenable it ... at some point in the future.

brideck:
Not sure where fairness comes into it. With the test excluded, both implementations can pass the TCK without breaking anyone. The correct implementation can also have the satisfaction of actually being correct. This is in no way a new or unique scenario with Jakarta EE TCKs.

CDI 4.0 is a finished spec that implementations have already certified against. Including this change in a service release would effectively invalidate those prior certifications. CDI 3.0 is equally wrong in the TCK and implementation, but you wouldn't make this change there, would you?

PR #445, opened 2023-03-09
Selenium Fix for Chromium v111+

brideck:
The new version of Chrome (111, released Mar 7th) has included some security updates that disable the websocket connection between the web driver and the browser. This fix is needed to continue testing with current versions of Chrome. The alternative would be to only test with versions no newer than 110, but Google does not make it easy to find back-level versions.

For a full discussion on the change, see: https://groups.google.com/g/chromedriver-users/c/xL5-13_qGaA


Emily-Jiang (approved):
LGTM

PR #449, opened 2023-04-19
@PostConstruct method on interceptor SheepInterceptor should have InvocationContext parameter

### LINKED ISSUE #448: Post construct interceptor missing an invocation context parameter
The `org.jboss.cdi.tck.interceptors.tests.contract.lifecycleCallback.SheepInterceptor` declares a `@PostConstruct` method which violates interceptor spec requirements.

Namely in [this section](https://jakarta.ee/specifications/interceptors/2.1/jakarta-interceptors-spec-2.1.html#interceptor-methods-for-lifecycle-event-callbacks), the following is stated:

> Lifecycle callback interceptor methods declared in an interceptor class or superclass of an interceptor class must have one of the following signatures:
>* `void <METHOD>(InvocationContext)`
>* Object <METHOD>(InvocationContext)

Here is a [link](https://github.com/jakartaee/cdi-tck/blob/4.0.9/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/contract/lifecycleCallback/SheepInterceptor.java#L32-L35) to the current method declaration which lacks the `InvocationContext` parameter.

The only test using this interceptor asserts that this callback *does not* get invoked which is probably why it didn't get noticed earlier.

~~EDIT: I intentionally didn't mark the issue as a challenge but if we encounter an impl having issues with it, it is certainly an option.~~
After some though, I added the challenge label.

manovotn:
Fixes #448

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/contract/lifecycleCallback/SheepInterceptor.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/contract/lifecycleCallback/SheepInterceptor.java
@@ -30,8 +30,13 @@ public class SheepInterceptor {
     private static boolean aroundInvokeCalled = false;
 
     @PostConstruct
-    public void postConstruct() {
+    public void postConstruct(InvocationContext ctx) {
         postConstructCalled = true;
+        try {
+            ctx.proceed();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
     }
 
     @AroundInvoke


Ladicek (approved):
The exception catching and rethrowing looks ugly and I personally think that declaring `throws Exception` would be fine, because the Interceptors spec makes a distinction between "declaring to throw checked exceptions" and "throwing checked exceptions", but it's not a big deal.

Ladicek:
Since #448 is not a challenge, we can't merge this into 4.0 TCK I believe.

I also believe this would be a valid challenge, but I'll leave that to you to decide :-)

manovotn:
> The exception catching and rethrowing looks ugly and I personally think that declaring `throws Exception` would be fine, because the Interceptors spec makes a distinction between "declaring to throw checked exceptions" and "throwing checked exceptions", but it's not a big deal.

It doesn't make a difference as the interceptor is meant not to be invoked in the first place :shrug: 

> I also believe this would be a valid challenge, but I'll leave that to you to decide :-)

Fine, let's do it.

PR #456, opened 2023-05-01
fix InterceptorDefinitionTest.testInterceptorsImplementInterceptorInterface

### LINKED ISSUE #455: org.jboss.cdi.tck.tests.interceptors.definition.InterceptorDefinitionTest#testInterceptorsImplementInterceptorInterface uses internal type
This test uses a helper class `HierarchyDiscovery` to find the set of interface types that an interceptor metadata class (implementation of `jakarta.enterprise.inject.spi.Interceptor`) implements and then verifies that a parameterized type `Interceptor` is among them. It expects that a `ParameterizedTypeImpl` is used to represent a parameterized type, which is an internal implementation detail of `HierarchyDiscovery`. If the the interceptor metadata class doesn't have a generic signature (which happens for our implementation, ArC, that generates the implementation at compile time), the parameterized type of `Interceptor` isn't the TCK's internal `ParameterizedTypeImpl`, but the JDK's `ParameterizedTypeImpl`. They both implement `java.lang.reflect.ParameterizedType`, which is something the test should use in my opinion.

Ladicek:
This test uses an internal `ParameterizedTypeImpl` class, which isn't guaranteed to be present in the set of interface types implemented by an interceptor metadata class. The JDK implementation of `ParameterizedType` may be present as well.

This commit uses the public interface `ParameterizedType` to perform this test.

Fixes #455

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/interceptors/definition/InterceptorDefinitionTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/interceptors/definition/InterceptorDefinitionTest.java
@@ -34,9 +34,7 @@ import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.cdi.tck.AbstractTest;
 import org.jboss.cdi.tck.shrinkwrap.WebArchiveBuilder;
 import org.jboss.cdi.tck.util.HierarchyDiscovery;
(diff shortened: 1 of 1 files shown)


Ladicek:
Draft until #455 is accepted.

manovotn (approved):

PR #458, opened 2023-05-01
move DestroyForSameCreationalContext2Test to the "full" part of the TCK

### LINKED ISSUE #457: org.jboss.cdi.tck.tests.context.DestroyForSameCreationalContext2Test isn't marked as "full" and uses session context and custom `Contextual`
The test `DestroyForSameCreationalContext2Test` repeats the exact same 2 assertions against a session context and a request context. To do that, it also uses a custom implementation of the `Contextual` interface. This test isn't marked as "full", and isn't a `full` package, because it makes sense for CDI Lite as well -- except CDI Lite implementations are not required to provide the session context and don't necessarily have to support custom implementations of `Contextual`.

I believe we should move the test to the `full` package.

Ladicek:
Fixes #457


Ladicek:
Draft until #457 is accepted.

Azquelt (approved):

PR #460, opened 2023-05-04
avoid stereotype annotations on injection points in InjectionPointTest

### LINKED ISSUE #459: org.jboss.cdi.tck.tests.lookup.injectionpoint.InjectionPointTest uses stereotype annotations on injection points
Two classes used in this test, `FieldInjectionPointBean` and `TransientFieldInjectionPointBean`, use a stereotype annotation on an injection point: 

- https://github.com/jakartaee/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/tests/lookup/injectionpoint/FieldInjectionPointBean.java#L32
- https://github.com/jakartaee/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/tests/lookup/injectionpoint/TransientFieldInjectionPointBean.java#L32

This is rather confusing. We have seen people doing this by mistake, so our implementation, ArC, actually rejects such code under certain circumstances.

Since the test verifies something completely different and the usage of stereotypes is unrelated to the test's purpose, I believe the annotation on the injection points mentioned above should be just a regular annotation.

Ladicek:
Fixes #459


Ladicek:
Draft until #459 is accepted.

impl/src/main/java/org/jboss/cdi/tck/tests/lookup/injectionpoint/TransientFieldInjectionPointBean.java
```diff
@@ -29,7 +29,7 @@
 @RequestScoped
 public class TransientFieldInjectionPointBean {
     @Inject
-    @AnimalStereotype
+    @Animal
```
manovotn, line 32: Is there a point in having any annotation here? The related test only checks the metadata to see if the IP is transient.

Ladicek: Not really, that annotation isn't used anywhere. I can remove it too.

Ladicek: Done.

manovotn (approved):

PR #462, opened 2023-05-11
fix test that interceptors are dependent instances of intercepted beans

### LINKED ISSUE #454: DependentContextTest#testDependentScopedInterceptorsAreDependentObjectsOfBean() does not test the spec assertion
The [`org.jboss.cdi.tck.tests.context.dependent.DependentContextTest#testDependentScopedInterceptorsAreDependentObjectsOfBean()`](https://github.com/jakartaee/cdi-tck/blob/4.1.0.Alpha3/impl/src/main/java/org/jboss/cdi/tck/tests/context/dependent/DependentContextTest.java#L446-L463) test method does not test the [linked spec assertion](https://github.com/jakartaee/cdi-tck/blob/4.1.0.Alpha3/impl/src/main/resources/tck-audit-cdi.xml#L2822-L2826):

> Instances of interceptors or decorators with scope `@Dependent` are dependent objects of the bean instance they decorate.

It merely asserts that if `CreationalContext#release()` is called then the `TransactionalInterceptor#destroy()` callback is invoked because the `AccountTransaction` bean is destroyed. And even this assert is questionable because the test should rather call `Bean#destroy()` instead; `CreationalContext#release()` should destroy the dependent objects and not the bean itself.

In fact, this assertion should be `testable=false` because there's no reasonable way to test it (an interceptor may not declare its own `@PreDestroy` callback).

I'd recommend to mark the assertion as `testable=false` and remove the test method completely.

Ladicek:
The test previously used a `@PreDestroy` callback declared on the interceptor to prove that the interceptor is destroyed when the intercepted bean is destroyed, which is incorrect. Interceptors cannot declare lifecycle callbacks for themselves, the `@PreDestroy` callback actually interposed on the destruction of the intercepted bean.

This commit uses a new dependent-scoped bean that is injected into the interceptor, which makes it a dependent instance of the interceptor. The new bean has a `@PreDestroy` callback for itself, which means we can verify that the new bean is destroyed when the intercepted bean is destroyed. Such chain of dependent instance destruction proves that the interceptor is indeed a dependent instance of the intercepted bean.

Fixes #454

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/context/dependent/DependentContextTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/context/dependent/DependentContextTest.java
@@ -414,7 +414,13 @@ public class DependentContextTest extends AbstractTest {
     @Test
     @SpecAssertion(section = DEPENDENT_OBJECTS, id = "aa")
     public void testDependentScopedInterceptorsAreDependentObjectsOfBean() {
-        TransactionalInterceptor.destroyed = false;
+        // since interceptors can't declare `@PreDestroy` callbacks for themselves, we inject
+        // another dependent-scoped bean into the interceptor and add a `@PreDestroy` callback
+        // there -- this dependency will be a dependent instance of the interceptor, and so
+        // if destroying the intercepted bean will destroy the other bean, we have a proof
+        // that the interceptor was also destroyed
+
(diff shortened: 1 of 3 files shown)


Ladicek:
Draft until #454 is accepted.

manovotn (approved):
LGTM

PR #452, opened 2023-04-26
avoid custom CreationalContext and Contextual implementations in the TCK

Ladicek:
On a small number of places, the CDI TCK creates its own implementations of `CreationalContext` and `Contextual`. These are extremely bare APIs that aren't sufficient for internal needs of CDI implementations or for interoperability. All CDI implementations that I know expose internally a much richer API and generally expect that their `CreationalContext` and `Contextual` implementations are used. It is possible to bend the implementations to accept a `CreationalContext` or `Contextual` of outside provenance, but I believe there's exactly nothing in the CDI specification that makes this mandatory.

Instead, I believe the CDI TCK should:

1. avoid using its own `CreationalContext` and `Contextual` implementations;
2. expose SPIs that lets implementations supply their own `CreationalContext` and `Contextual` implementations, augmented with inspection capabilities for the tests that need it.

This PR does exactly that.

PR #463, opened 2023-05-18
avoid custom creational context implementations

### LINKED ISSUE #453: Custom CreationalContext used by TCK is not required to be supported
On several places, TCK tests use a custom `CreationalContext` implementation - `org.jboss.cdi.tck.util.MockCreationalContext`.

However, the specification doesn't state that arbitrary custom creational contexts should be supported in general nor is any test asserting that claim.
Plus it's not realistic either as the API is very minimal and both impls I am familiar with need a much richer internal API to handle CC and, in turn, expect that any CC they operate on is an instance of their internal implementation.

Now, this custom CC is used in two ways:

1. In tests where there is need to create any instance of CC
  1.1 An example can be seen [here](https://github.com/jakartaee/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/tests/context/GetFromContextualTest.java#L54)
  1.2 These can be easily replaced by `BeanManager#createCreationalContext()` without changing what the test does - as such these can be altered even for 4.x releases without need for exclusion
2. In tests that need to verify what's happening in the CC; basically an introspection of how the CC should work
  2.1 An example is [this test](https://github.com/jakartaee/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/tests/implementation/simple/lifecycle/SimpleBeanLifecycleTest.java#L95) 
  2.2 We can replace these test with new TCK SPI which introduces and inspectable CC provided by the impl and therefore still being based on their internal impl. _This doesn't change the nature of what is tested, it merely let's the integrator provide an instance of CC which the TCK can inspect and verify._

@Ladicek already created a proposal for the TCK SPI that could be used in future version of TCKs (minor/major update); see https://github.com/jakartaee/cdi-tck/pull/452

For current version of TCKs linked to CDI 4.0, I'd suggest we exclude any tests that rely on introspection of the CC. WDYT?

Ladicek:
This commit removes an unused class `CreationalContextImpl` and replaces most usages of `MockCreationalContext` with `BeanManager.createCreationalContext()` (because those places don't really use the `MockCreationalContext` for verifying anything). The 2 remaining usages of `MockCreationalContext` are added to the canonical exclude list with the plan to enable them again in subsequent release, after migrating to a TCK SPI (#452).

Fixes #453


manovotn (approved):

manovotn:
This is exactly what I meant in https://github.com/jakartaee/cdi-tck/issues/453#issue-1684959234
The secondary part can then be addressed by https://github.com/jakartaee/cdi-tck/pull/452 but needs to go exclusively into next minor/major (whichever we do).

PR #465, opened 2023-05-24
Change InvocationContextTest#testGetTargetMethod() to avoid using recursive interception

### LINKED ISSUE #464: InvocationContextTest#testGetTargetMethod() relies on undefined behavior
I believe the TCK test [`InvocationContextTest#testGetTargetMethod()`](https://github.com/jakartaee/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/contract/invocationContext/InvocationContextTest.java#L47-L50) is incorrect in how it tests given behavior.
It currently relies on recursive interception to verify that `InvocationContext#getTarget()` returns the same instance of a bean. Namely the test invokes `simpleBean.getId()` and the around invoke interceptor then uses `InvocationContext#getTarget()` to obtain an instance of `SimpleBean` and invokes `getId()` again.

Correct me if I am wrong, but I couldn't find any evidence in CDI or Interceptors specs that would imply recursive interception is supported or even supposed to work. Now, I am not for or against that feature but if we want it, it should have it's own clear definition in either spec and its dedicated TCK test.

Note that the exact same test goal can be achieved without use of recursive interception (for instance a static field in the interceptor and comparing the two instances?). Therefore, we should be able to tweak the test to avoid recursive interception invocations without losing any coverage.

manovotn:
Fixes #464 

This PR changes the test in question to test return value of `InvocationContetx#getTarget()` without relying on recursive interception. Verified with Weld and Arc.

PR is a draft until linked issue gets the `accepted` label.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/contract/invocationContext/Interceptor1.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/contract/invocationContext/Interceptor1.java
@@ -25,21 +25,16 @@ import jakarta.interceptor.InvocationContext;
 @Binding1
 @Priority(100)
 public class Interceptor1 {
-    private static boolean getTargetOK = false;
+    private static SimpleBean target;
 
     @AroundInvoke
     public Object aroundInvoke(InvocationContext ctx) throws Exception {
-        SimpleBean target = (SimpleBean) ctx.getTarget();
-        int id1 = target.getId();
-        int id2 = (Integer) ctx.proceed();
-
-        if (id1 == id2) {
-            getTargetOK = true;
-        }
-        return id1;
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/contract/invocationContext/Interceptor1.java
```diff
@@ -26,20 +26,16 @@
 @Priority(100)
 public class Interceptor1 {
     private static boolean getTargetOK = false;
```
Ladicek, line 28: This could also be removed I guess?

manovotn: Done

Ladicek (approved):

PR #466, opened 2023-05-24
Clarify requirements of the Contexts SPI interface

### LINKED ISSUE #447: ConditionalObserverTest#testConditionalObserverMethodNotInvokedIfNoActiveContext
Hello,

I recently came across the [test in title](https://github.com/jakartaee/cdi-tck/blob/4.0.9/impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/conditional/ConditionalObserverTest.java#L101-L123) and I think that it has unfounded expectations.
Namely, the test does the following steps:

1. Create instance of a bean that holds observer method with reception `IF_EXISTS` 
2. Manually deactivate context of the bean created in 1. (deactivate, not destroy)
3. Fire an event to verify that observer of the bean from step 1. isn't notified
4. Manually re-activate context of the bean from step 1.
5. Fire a new event and verify that the exact same bean instance from step 1. now got notified

Now, what I find dubious is the step 5. where the test assumes that after reactivating context, the very same bean instance is again present in that context, hence OM with `IF_EXIST` reception will go through.
Note that activation and deactivation of contexts happens through [CDI TCK SPI `Contexts`](https://github.com/jakartaee/cdi-tck/blob/4.0.9/api/src/main/java/org/jboss/cdi/tck/spi/Contexts.java#L36-L48). This doesn't have much documentation apart from its javadoc and a mention that it exists in TCK docs.

I couldn't find any part of spec that would support this test in terms of deactivation and subsequent activation implicating that the context state is renewed with the same set of contextual references.

Note 1: The specific part of spec that the test checks (denoted by `@SpecAssertion` on it) is checked by steps 1 to 4 so the test can be easily altered.
Note 2: Both, Weld and Arc (Quarkus) are able to pass this test, so this is nothing blocking, I am mostly curious whether I am missing something in the spec that would warrant this test.

manovotn:
Fixes #447 

Documents the behavior of the `org.jboss.cdi.tck.spi.Contexts` SPI in terms of context re-activation, aligning it with how it's used in tests.
An example of an existing test relying on exactly this behavior is `ConditionalObserverTest#testConditionalObserverMethodNotInvokedIfNoActiveContext()`

### DIFF
--- a/api/src/main/java/org/jboss/cdi/tck/spi/Contexts.java
+++ b/api/src/main/java/org/jboss/cdi/tck/spi/Contexts.java
@@ -34,14 +34,20 @@ public interface Contexts<T extends Context> {
     public static final String PROPERTY_NAME = Contexts.class.getName();
 
     /**
-     * Sets the specified context as active
+     * Sets the specified context as active.
+     * <p>
+     * The set of existing contextual instances of the context is preserved
+     * across invocations of {@link #setActive(Context)} and {@link #setInactive(Context)}.
      *
      * @param context The context to set active
      */
     public void setActive(T context);
 
     /**
-     * Sets the specified context as inactive
+     * Sets the specified context as inactive.
+     * <p>
+     * The set of existing contextual instances of the context is preserved
+     * across invocations of {@link #setActive(Context)} and {@link #setInactive(Context)}.
      *
      * @param context The context to set inactive
      */


api/src/main/java/org/jboss/cdi/tck/spi/Contexts.java
```diff
@@ -34,14 +34,22 @@
     public static final String PROPERTY_NAME = Contexts.class.getName();
 
     /**
-     * Sets the specified context as active
+     * Sets the specified context as active.
+     * <p>
+     * If the context was previously set as inactive via {@link  #setInactive(Context)}, this method is expected to
+     * restore the state of the context, including all contextual instances that were previously in it.
+     * <p>
+     * If the context wasn't previously set as inactive, a new context is created and activated.
```
Ladicek, line 42: That can't be true. The `setActive` method accepts a context object, "a new context" can't be created.

Also, I'd avoid terms like "context state", as that's not a term defined by the specification. I believe something like this would be enough:

```java
/**
 * Sets the specified context as active.
 * <p>
 * The set of existing contextual instances of the context is preserved
 * across invocations of {@link #setActive(Context)} and {@link #setInactive(Context)}.
 *
 * @param context The context to set active
 */
public void setActive(T context);

/**
 * Sets the specified context as inactive.
 * <p>
 * The set of existing contextual instances of the context is preserved
 * across invocations of {@link #setActive(Context)} and {@link #setInactive(Context)}.
 *
 * @param context The context to set inactive
 */
public void setInactive(T context);
```

WDYT?

manovotn: > That can't be true. The setActive method accepts a context object, "a new context" can't be created.

You're right, that's poor wording. I meant to convey that the set of instances won't be there in any other case than when you previously set that context as inactive.

> * The set of existing contextual instances of the context is preserved
> * across invocations of {@link #setActive(Context)} and {@link #setInactive(Context)}.

Definitely better; but shouldn't we imply that the contextual instances are preserved only if you do `setInactive()` -> `setActive()`? Note that there is also destroy method plus you could attempt to invoke `setActive()` repeatedly after you destroy a contextual instance. But I guess I am just overthinking it :)

manovotn: Updated the text per your suggestion

Ladicek (approved):

PR #469, opened 2023-05-29
Change InterceptorBindingInheritanceTest to avoid using recursive interception

### LINKED ISSUE #468: InterceptorBindingInheritanceTest relies on undefined behavior
Similarly to #464, two tests in `InterceptorBindingInheritanceTest` rely on recursive interception being prevented by the container, which is not defined properly anywhere and is unrelated to the stated purpose of the test (which is inheritance of interceptor binding annotations).

It is relatively straightforward to modify the test to not trigger recursive interception, which is what I propose we do.

Ladicek:
Fixes #468


Ladicek:
Draft until the challenge is accepted.

manovotn (approved):

manovotn:
I'm going to fast forward this into main as it has both, the labels and the approval and I am about to cut a TCK release (as I promised to do during last CDI call we had).
I'm happy to have further discussion around #468 and #464 if there is need for it.

PR #471, opened 2023-06-02
Correct the deployment setup of EmptyBeansXmlDiscoveryTest

### LINKED ISSUE #470: EmptyBeansXmlDiscoveryTest has incorrect deployment configuration
In my attempt to update Arquillian deps in Weld(https://github.com/weld/core/pull/2846), I noticed that `EmptyBeansXmlDiscoveryTest` started failing.
A recent fix in Arquillian (https://github.com/arquillian/arquillian-core/pull/440) correctly exposes erroneous setup of this test which was previously falsely passing and would pass regardless of what would be in the test.

Underlying cause is a `NoClassDefFound` error and further digging showed that the test deployment lacks the `AbstractTest` superclass (and possibly other classes).
Said test uses a manual setup via `ShrinkWrap.create(WebArchive.class).addClasses(...)` (beucase it needed a truly empty beans.xml added) whereas other TCK tests use `WebArchiveBuilder()` API. Turns out `WebArchiveBuilder` does a bunch of things for the test automagically, including adding `AbstractTest` class into the deployment.

The test can be easily rewritten to make use of the `WebArchiveBuilder` which solves the issue and should make it (correctly) pass on al Arq. versions. I have the changes locally and will submit a PR.

Since this problem is a combination of testng and Arq. core version, not all impls and test setups are affected - especially since you can do a lot of tweaks in Arq. adapters that each impl is using. Nonetheless, the test setup is still wrong. I wonder if perhaps Liberty is also seeing this if they update Arq. core to `1.7.0.Final`, maybe @Azquelt knows?

manovotn:
Fixes #470 

Changes how the test deployment is created - doesn't change the functionality or contents of the test.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/deployment/discovery/EmptyBeansXmlDiscoveryTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/deployment/discovery/EmptyBeansXmlDiscoveryTest.java
@@ -7,20 +7,21 @@ import jakarta.enterprise.inject.Instance;
 import jakarta.enterprise.inject.spi.CDI;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.cdi.tck.AbstractTest;
-import org.jboss.shrinkwrap.api.ShrinkWrap;
-import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.cdi.tck.shrinkwrap.WebArchiveBuilder;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.jboss.test.audit.annotations.SpecVersion;
 import org.testng.annotations.Test;
 
 /**
  * Tests that a singular archive with empty beans.xml results in annotated discovery mode.
  */
+@SpecVersion(spec = "cdi", version = "4.0")
 public class EmptyBeansXmlDiscoveryTest extends AbstractTest {
 
     @Deployment
     public static WebArchive createTestArchive() {
-        return ShrinkWrap.create(WebArchive.class).addClasses(SomeAnnotatedBean.class, SomeUnannotatedBean.class)
-                .addAsWebInfResource(new StringAsset(""), "beans.xml");
+        return new WebArchiveBuilder().withTestClassPackage(EmptyBeansXmlDiscoveryTest.class)
+                .withBeansXml("beans.xml").build();
     }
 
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/deployment/discovery/EmptyBeansXmlDiscoveryTest.java
```diff
@@ -7,20 +7,21 @@
...
  */
+@SpecVersion(spec = "cdi", version = "4.0")
 public class EmptyBeansXmlDiscoveryTest extends AbstractTest {
 
     @Deployment
     public static WebArchive createTestArchive() {
-        return ShrinkWrap.create(WebArchive.class).addClasses(SomeAnnotatedBean.class, SomeUnannotatedBean.class)
-                .addAsWebInfResource(new StringAsset(""), "beans.xml");
+        return new WebArchiveBuilder().withTestClassPackage(EmptyBeansXmlDiscoveryTest.class)
+                .withBeansXml("beans.xml").build();
```
Ladicek, line 24: Shouldn't this be `withBeansXml("")`?

manovotn: Maybe both work? If you look at the usages of this method, all other tests use the syntax I have here

Ladicek: Ah I get it now. What you have is correct.

It's just that the `withBeansXml` javadoc is super-confusing and its source code, without looking further, doesn't help either. The `String` parameter is a path to the classpath resource. It would seem you don't have to specify that, because the javadoc says:

```
Add <code>beans.xml</code> located in src/main/resource/{testPackagePath}.
```

but you do. The `beans.xml` from the javadoc refers to where the file is _placed into_, not where it is _read from_.

It should read something like this IMHO:

```
Add a file with given {@code beansXml} name located in src/main/resource/{testPackagePath} to the archive as <code>beans.xml</code>.
```

Sorry for the noise.

Emily-Jiang (approved):
LGTM! The fix is pretty minor and the correct way to fix this. If some runtime challenges this, we can exclude this test and do a subsequent release.

Ladicek (approved):

PR #477, opened 2023-07-12
Avoid using collection as bean type in ManagedBeanTypesTest

### LINKED ISSUE #476: ManagedBeanTypesTest not compatible with JDK 21
`ManagedBeanTypesTest` in the CDI TCK Passes on JDK 17, but on JDK 21 (ea30) contains an extra collection, resulting in the assertion error:

```
Set 

[
class java.lang.Object, 
class org.jboss.cdi.tck.tests.definition.bean.types.Flock, 
java.util.List<org.jboss.cdi.tck.tests.definition.bean.types.Vulture<java.lang.Integer>>, 

interface java.util.SequencedCollection<org.jboss.cdi.tck.tests.definition.bean.types.Vulture<java.lang.Integer>>, 

interface java.util.Collection<org.jboss.cdi.tck.tests.definition.bean.types.Vulture<java.lang.Integer>>, 
interface java.lang.Iterable<org.jboss.cdi.tck.tests.definition.bean.types.Vulture<java.lang.Integer>>
] (6) 

does not match array 

[
class java.lang.Object, 
class org.jboss.cdi.tck.tests.definition.bean.types.Flock, 
java.util.List<org.jboss.cdi.tck.tests.definition.bean.types.Vulture<java.lang.Integer>>, 

java.util.Collection<org.jboss.cdi.tck.tests.definition.bean.types.Vulture<java.lang.Integer>>, 
java.lang.Iterable<org.jboss.cdi.tck.tests.definition.bean.types.Vulture<java.lang.Integer>>
] (5)
```

manovotn:
Fixes #476 

IIUIC, this should fix the problem while retaining the test goal (which is bean types for nested generics).
Is that right @arjantijms ?


Ladicek:
I would avoid `Iterable` too, even if it seems fairly safe.

But yes, this is a proper way to fix this.

impl/src/main/java/org/jboss/cdi/tck/tests/definition/bean/types/Gathering.java
```diff
@@ -0,0 +1,4 @@
+package org.jboss.cdi.tck.tests.definition.bean.types;
+
+public interface Gathering<T> extends Iterable<T>{
```
Ladicek, line 3:
```suggestion
public interface Gathering<T> extends Iterable<T> {
```

arjantijms:
Looks good, same remark about `Iterable`.

manovotn:
> I would avoid `Iterable` too, even if it seems fairly safe.

> Looks good, same remark about `Iterable`.

Alright, alright, my laziness got the better of me :)
I've updated the PR to only use custom interface structure.

arjantijms (approved):

Ladicek (approved):

Ladicek:
Do we treat this (or rather #476) as a challenge?

manovotn:
> But just like an implementation detail was updated for the Authentication TCK that didn't effect the test goal itself (which is just testing generics in this case), it could be a general update. Thoughts?

I am not sure we are allowed to put anything but necessary changes and challenge resolutions to 4.0.x releases now.
It's probably safer to leave this for 4.1 - that being said, we should consider just moving main branch to that and creating a separate 4.0 branch for any further maintenance releases. Most work going into CDI now will be related to 4.1 anyway

Ladicek:
I added the 4.1 milestone.

Good point about branching @manovotn though. I want to submit the PR for TCK tests for `InvocationContext.getInterceptorBindings()`, and there are some other outstanding PRs for 4.1 already. So I agree with moving `master` to 4.1.

PR #479, opened 2023-07-17
Add tests for InvocationContext.getInterceptorBindings()

Ladicek:
TCK tests for https://github.com/jakartaee/interceptors/pull/99

Draft because there's a temporary profile to enable using Jakarta staging repository. Once Interceptors 2.2.0-RC1 hits Maven Central, I'll remove that and this PR will be ready.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/contract/invocationContext/AroundConstructBinding1.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/interceptors/tests/contract/invocationContext/AroundConstructBinding1.java
@@ -1,21 +1,22 @@
 package org.jboss.cdi.tck.interceptors.tests.contract.invocationContext;
 
+import jakarta.enterprise.util.AnnotationLiteral;
+import jakarta.interceptor.InterceptorBinding;
+
 import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
-import jakarta.enterprise.util.AnnotationLiteral;
-import jakarta.interceptor.InterceptorBinding;
-
+import static java.lang.annotation.ElementType.CONSTRUCTOR;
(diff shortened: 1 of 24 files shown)


Ladicek:
This would be ready now, I believe, but let's keep it as a draft until https://github.com/jakartaee/interceptors/pull/103 is merged.

manovotn:
Moved from draft to PR as the linked PR was already approved.
Also main branch of TCKs is now set for 4.1 so we should be able to get this in.

Ladicek:
I just realized we didn't add a test for interceptor bindings added through `InterceptionFactory`. @manovotn would you have some time to add that? If not, I can try.

manovotn:
> I just realized we didn't add a test for interceptor bindings added through `InterceptionFactory`. @manovotn would you have some time to add that? If not, I can try.

Ok, I'll make a note and send a PR

manovotn:
> Adding an interceptor binding to a method or type using an extension

True, although there is little chance this wouldn't work assuming the interceptor itself is bound correctly (i.e. if that binding was registered, I'd be very surprised if it weren't visible here)

> [A binding type with additional bindings](https://jakarta.ee/specifications/interceptors/2.1/jakarta-interceptors-spec-2.1#interceptor_binding_types_with_additional_interceptor_bindings) - not sure if the additional binding should be returned here or not

Good point and those should be returned as well because of the [javadoc](https://github.com/jakartaee/interceptors/blob/master/api/src/main/java/jakarta/interceptor/InvocationContext.java#L145-L159) wording of the method:

>     * All interceptor binding annotations are returned, including interceptor binding annotations
>     * that associate interceptors of a different interceptor method type, as well as interceptor
>     * binding annotations that associate no interceptor.

> I can make a PR for the second one, and I'd be happy to do the first one as well if we can determine the correct behaviour.

:+1:

PR #486, opened 2023-10-10
Specify event that triggers exception

jbescos:
`ObserverExceptionRethrownTest#testNonTransactionalObserverThrowsNonCheckedExceptionIsRethrown` verifies that one exception happens when firing one event.

 In the case of Helidon, this exception is thrown on deployment `ObserverExceptionRethrownTest#createTestArchive` because Helidon container fires 2 events on deployment. One to notify when beans are going to be deployed, and other when beans are deployed.

Proposed solution: `TeaCupPomeranian` will evaluate the event content, to throw the exception only when we want to throw it.


manovotn:
> Proposed solution: TeaCupPomeranian will evaluate the event content, to throw the exception only when we want to throw it.

I'd rather use a custom event payload (some `Foobar` instead of `String`) which I suppose will solve your issue as well?

Also, do you need this fixed in any existing TCKs? Such as TCKs for CDI 4?
If so, we would need to have a challenge for that and another PR for that branch (`master` is for CDI 4.1., the upcoming version).

jbescos:
I have modified the PR to a custom event payload. That solves it too, thanks for the point.

I need it for CDI 4.0. I am going to create a challenge for that.

Thank you.

manovotn (approved):

PR #496, opened 2023-11-07
Relax the scope requirement for public fields

Ladicek:
This is a TCK counterpart to the [specification change](https://github.com/jakartaee/cdi/pull/698) that allows all pseudo-scoped beans to have public fields. It updates the text in `tck-audit-cdi.xml` files, fixes naming in the relevant tests and adds a test that a `@Singleton` bean can have a public field.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/implementation/simple/definition/SimpleBeanDefinitionTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/implementation/simple/definition/SimpleBeanDefinitionTest.java
@@ -118,4 +118,10 @@ public class SimpleBeanDefinitionTest extends AbstractTest {
         Assert.assertEquals(getContextualReference(Tiger.class).name, "pete");
     }
 
+    @Test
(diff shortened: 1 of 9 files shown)


manovotn:
@Ladicek shouldn't we add a test that verifies this for `@Singleton`?

Ladicek:
We could, but it's not like we can ever be exhaustive. For normal scopes, we only test `@RequestScoped`, even though we could also test `@ApplicationScoped` easily :shrug:

manovotn:
> We could, but it's not like we can ever be exhaustive. For normal scopes, we only test `@RequestScoped`, even though we could also test `@ApplicationScoped` easily 🤷

True, I only mentioned this as it is a notable difference from previous version of specification.
I'll leave it up to you

manovotn (approved):

Ladicek:
I could add a test for `@Singleton` as well, but `@Singleton` is not a bean defining annotation, so I would have to create a stereotype, and I'm just lazy...

OK, I will do it :-)

Ladicek:
Done.

manovotn (approved):
Thank you and sorry for the bother :)

PR #497, opened 2023-11-08
fix InvocationContextTest and InterceptorBindingInheritanceTest

Ladicek:
This is to cater for implementations that implement interception by proxying, in which case the contextual instance of a bean is an interception proxy, while `InvocationContext.getTarget()` returns the proxied object (an "actual" instance of the bean).

Stems from recent discussion in #468


manovotn (approved):
LGTM.
I've also verified this test with latest Weld and it passes just fine.

PR #515, opened 2023-12-08
synchronize exclusion lists across `impl` and `web` + remove `tck-tests-previous.xml`

Ladicek:


Ladicek:
Added a commit (here and in #514) to run CI for PRs against all branches.

manovotn (approved):
Looks good and thanks for checking on the CI bit as well!

PR #520, opened 2024-01-04
Test obtaining interceptor bindings along with interception factory

manovotn:
A follow up on https://github.com/jakartaee/cdi-tck/pull/479 which didn't have any test that would leverage interception factory as a source of bindings.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/invocationContext/ProducerBean.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/invocationContext/ProducerBean.java
@@ -1,0 +1,24 @@
+package org.jboss.cdi.tck.tests.full.interceptors.contract.invocationContext;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Produces;
+import jakarta.enterprise.inject.spi.InterceptionFactory;
+import jakarta.enterprise.inject.spi.configurator.AnnotatedTypeConfigurator;
+
+@ApplicationScoped
+public class ProducerBean {
+
+    @Produces
+    @ApplicationScoped
+    public Product createProduct(InterceptionFactory<Product> interceptionFactory) {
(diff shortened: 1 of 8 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/invocationContext/ProducerBean.java
```diff
@@ -0,0 +1,19 @@
...
+@ApplicationScoped
+public class ProducerBean {
+
+    @Produces
+    @ApplicationScoped
+    public Product createProduct(InterceptionFactory<Product> interceptionFactory) {
+        interceptionFactory.configure().add(ProductInterceptorBinding1.Literal.INSTANCE).filterMethods(m -> m.getJavaMember().getName().equals("ping"))
+                .findFirst().get()
+                .add(ProductInterceptorBinding2.Literal.INSTANCE)
+                .add(ProductInterceptorBinding3.Literal.INSTANCE);
```
Ladicek, line 16: Wouldn't it be better to add only `ProductInterceptorBinding2` to `ping` and have another method to which we would add `ProductInterceptorBinding3`?

Ladicek (approved):
LGTM, just a minor note.

impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/invocationContext/ProducerBean.java
manovotn: @Ladicek I've updated the PR to use two different interceptors and two methods that have both different sets of bindings

Ladicek (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/full/interceptors/contract/invocationContext/ProducerBean.java
Ladicek: Thanks!

PR #524, opened 2024-01-11
add more tests for InvocationContext.getInterceptorBindings()

Ladicek:
- interceptor binding inherited from a superclass
- method-level interceptor binding "overriding" a class-level interceptor binding of the same type

Follows up on #522


Ladicek:
CC @Azquelt

Azquelt (approved):

manovotn (approved):

PR #526, opened 2024-01-12
Test BeanContainer assignability API methods

### LINKED ISSUE #518: Create TCK coverage for programmatic access to assignability rules for observer methods and typesafe resolution
CDI issue - https://github.com/jakartaee/cdi/issues/498
CDI PR - https://github.com/jakartaee/cdi/pull/700

We don't need to go in-depth on the assignability itself as those are covered elsewhere.
Instead, we should mostly test some basic cases plus the fact that implicit qualifiers are added automatically.

Azquelt:
Tests for .isMatchingBean and .isMatchingEvent.

Fixes #518

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanEventAssignabilityTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanEventAssignabilityTest.java
@@ -1,0 +1,244 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.cdi.tck.tests.beanContainer.assignability;
+
+import static org.jboss.cdi.tck.cdi.Sections.BM_BEAN_EVENT_ASSIGNABILITY;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertThrows;
+import static org.testng.Assert.assertTrue;
+
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.Set;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.cdi.tck.AbstractTest;
+import org.jboss.cdi.tck.shrinkwrap.WebArchiveBuilder;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.jboss.test.audit.annotations.SpecAssertion;
+import org.jboss.test.audit.annotations.SpecVersion;
+import org.testng.annotations.Test;
+
+import jakarta.enterprise.inject.Any;
+import jakarta.enterprise.inject.Default;
+import jakarta.enterprise.inject.literal.NamedLiteral;
+import jakarta.enterprise.util.TypeLiteral;
+
+/**
+ * @author Andrew Rouse
+ */
+@SpecVersion(spec = "cdi", version = "4.1")
+public class BeanEventAssignabilityTest extends AbstractTest {
+
+    @Deployment
+    public static WebArchive createTestArchive() {
+        return new WebArchiveBuilder().withTestClassPackage(BeanEventAssignabilityTest.class).build();
+    }
+
+    @Test
+    @SpecAssertion(section = BM_BEAN_EVENT_ASSIGNABILITY, id = "aa")
+    @SpecAssertion(section = BM_BEAN_EVENT_ASSIGNABILITY, id = "af")
+    public void testBeanMatching() {
+        Set<Type> beanTypes = Set.of(MyBean.class, MyBeanInterface.class, Object.class);
+
+        // Basic checks with qualifiers
+        assertTrue(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(), MyBean.class, Set.of()),
+                "Bean did not match its own type");
+        assertFalse(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(), MyBean.class, Set.of(Qualifier1.Literal.INSTANCE)),
+                "Bean matched despite not having required qualifier");
+        assertTrue(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(Qualifier1.Literal.INSTANCE), MyBean.class, Set.of(Qualifier1.Literal.INSTANCE)),
+                "Bean did not match despite having required qualifier");
+        assertTrue(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(Qualifier1.Literal.INSTANCE, Qualifier2.Literal.INSTANCE), MyBean.class, Set.of(Qualifier1.Literal.INSTANCE)),
+                "Bean did not match despite having a superset of the required qualifiers");
+        
+        // Only bean types passed in should be considered
+        Set<Type> reducedBeanTypes = Set.of(MyBean.class);
+        assertTrue(getCurrentBeanContainer().isMatchingBean(reducedBeanTypes, Set.of(), MyBean.class, Set.of()),
+                "Bean did not match its own type");
+        assertFalse(getCurrentBeanContainer().isMatchingBean(reducedBeanTypes, Set.of(), MyBeanInterface.class, Set.of()),
+                "Bean matched MyBeanInterface despite it not being in bean types");
+        assertTrue(getCurrentBeanContainer().isMatchingBean(reducedBeanTypes, Set.of(), Object.class, Set.of()),
+                "Bean did not match when Object requested");
+        
+        // Qualifier annotations on bean type classes should not be considered
+        assertTrue(getCurrentBeanContainer().isMatchingBean(Set.of(MyQualifiedBean.class), Set.of(), MyQualifiedBean.class, Set.of()),
+                "MyQualifiedBean should match, qualifier on bean class should be ignored");
+        assertFalse(getCurrentBeanContainer().isMatchingBean(Set.of(MyQualifiedBean.class), Set.of(), MyQualifiedBean.class, Set.of(Qualifier1.Literal.INSTANCE)),
+                "MyQualifiedBean should not match, qualifier on bean class should be ignored");
+    }
+
+    @Test
+    @SpecAssertion(section = BM_BEAN_EVENT_ASSIGNABILITY, id = "ab")
+    public void testBeanMatchingDefaultQualifiers() {
+        Set<Type> beanTypes = Set.of(MyBean.class, MyBeanInterface.class, Object.class);
+        
+        assertTrue(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(Default.Literal.INSTANCE), MyBean.class, Set.of()),
+                "Bean with @Default should match when no qualifiers required");
+        
+        assertTrue(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(), MyBean.class, Set.of(Default.Literal.INSTANCE)),
+                "Bean with no qualifiers should match when @Default required");
+        
+        assertTrue(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(Any.Literal.INSTANCE), MyBean.class, Set.of()),
+                "Bean with explicit @Any should match when no qualifiers required");
+
+        assertTrue(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(NamedLiteral.of("foo")), MyBean.class, Set.of()),
+                "Bean with @Named should match when no qualifiers required");
+        
+        assertFalse(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(Qualifier1.Literal.INSTANCE), MyBean.class, Set.of()),
+                "Bean with @Qualifier1 should not match when no qualifiers required (@Default implied required)");
+        
+        assertTrue(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(), MyBean.class, Set.of(Any.Literal.INSTANCE)),
+                "Bean with no qualifiers should match when @Any required");
+        
+        assertTrue(getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(Qualifier1.Literal.INSTANCE), MyBean.class, Set.of(Any.Literal.INSTANCE)),
+                "Bean with @Qualifier1 should match when @Any required");
+    }
+
+    @Test
+    @SpecAssertion(section = BM_BEAN_EVENT_ASSIGNABILITY, id = "ac")
+    public void testBeanMatchingNullException() {
+        
+        Set<Type> beanTypes = Set.of(MyBean.class, MyBeanInterface.class, Object.class);
+        
+        assertThrows("Null bean type", IllegalArgumentException.class, () -> getCurrentBeanContainer().isMatchingBean(null, Set.of(), MyBean.class, Set.of()));
+        assertThrows("Null bean qualifiers", IllegalArgumentException.class, () -> getCurrentBeanContainer().isMatchingBean(beanTypes, null, MyBean.class, Set.of()));
+        assertThrows("Null required type", IllegalArgumentException.class, () -> getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(), null, Set.of()));
+        assertThrows("Null required qualifiers", IllegalArgumentException.class, () -> getCurrentBeanContainer().isMatchingBean(beanTypes, Set.of(), MyBean.class, null));
+    }
+
+    @Test
+    @SpecAssertion(section = BM_BEAN_EVENT_ASSIGNABILITY, id = "ad")
+    public void testBeanMatchingNonQualifiersException() {
+
+        Set<Type> beanTypes = Set.of(MyBean.class, MyBeanInterface.class, Object.class);
+
+        assertThrows("beanQualifiers annotation not a qualifier", IllegalArgumentException.class,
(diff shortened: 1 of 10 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanObserverAssignabilityTest.java
```diff
@@ -0,0 +1,187 @@
...
+                "Bean did not match despite having a superset of the required qualifiers");
+        
+        // Only bean types passed in should be considered
+        Set<Type> reducedBeanTypes = Set.of(MyBean.class);
+        assertTrue(getCurrentBeanContainer().isMatchingBean(reducedBeanTypes, Set.of(), MyBean.class, Set.of()),
+                "Bean did not match its own type");
+        assertFalse(getCurrentBeanContainer().isMatchingBean(reducedBeanTypes, Set.of(), MyBeanInterface.class, Set.of()),
+                "Bean matched MyBeanInterface despite it not being in bean types");
+        assertFalse(getCurrentBeanContainer().isMatchingBean(reducedBeanTypes, Set.of(), Object.class, Set.of()),
+                "Bean matched Object despite it not being in bean types");
```
Azquelt, line 72: Wasn't sure about this test.

Every bean has `Object` as a bean type. Should `Object` be implied if not passed in the `beanTypes` set?

Azquelt:
Also worth noting:
* The API that these tests depend on hasn't been released yet. I needed to change the `cdi.api.version` to `4.1.0-SNAPSHOT` and build the CDI API locally.
* These new API methods aren't implemented yet so I haven't been able to run the tests.

manovotn:
> These new API methods aren't implemented yet so I haven't been able to run the tests.

Should be fairly easy to do in Weld, I'll make a branch next week so that we can test it.

impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanObserverAssignabilityTest.java
manovotn: > Every bean has `Object` as a bean type. Should `Object` be implied if not passed in the `beanTypes` set?

Yes, that's always added.
I think there was even a TCK for `@Typed` that asserted that `Object` is added automatically.

ljnelson: (Lurking. Curious if this is generally true in the specification? I see that certainly for `@Typed` `Object.class` is implied, but if I make a synthetic bean and explicitly do not include `Object.class` in its `Set<Type>`...?)

manovotn: > For example, https://jakarta.ee/specifications/cdi/4.0/apidocs/jakarta.cdi/jakarta/enterprise/inject/spi/configurator/beanattributesconfigurator#types(java.util.Set) says nothing about Object.class being intrinsically added.)

Yes, I am aware of that. However, event a synth bean is a bean and as therefore the first link you posted should still apply. At least in my opinion :)
And I think that's also what Weld does; not sure if there is a TCK for it though 🤷

Azquelt: It might be helpful to clarify whether `Object` is automatically added with a note on the relevant methods.

For this method, it sounds like we think `Object` _should_ be automatically added to the set of bean types so this test should change to `assertTrue`?

Ladicek: If all synthetic beans are beans, and if all beans have the bean type `java.lang.Object`, then all synthetic beans have the bean type `java.lang.Object`. That is a straightforward implication in my eyes.

Now, whether `isMatchingBean` should automatically add `Object` to `beanTypes`, that's another question. It could, I don't think that would be harmful in any way, but the method specification should be explicit about it. Anyone wants to submit a PR?

The only possible downside is that someone might start asking "why don't we automatically scan for all superclasses and implemented interfaces?", and I wouldn't have a good answer :-)

impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanObserverAssignabilityTest.java
```diff
@@ -0,0 +1,187 @@
...
+        
+        // Qualifier annotations on bean type classes should not be considered
+        assertTrue(getCurrentBeanContainer().isMatchingBean(Set.of(MyQualifiedBean.class), Set.of(), MyQualifiedBean.class, Set.of()),
+                "MyQualifiedBean should match, qualifier on bean class should be ignored");
+        assertTrue(getCurrentBeanContainer().isMatchingBean(Set.of(MyQualifiedBean.class), Set.of(), MyQualifiedBean.class, Set.of(Qualifier1.Literal.INSTANCE)),
```
Ladicek, line 77: This should be `assertFalse`?

Azquelt: Correct :(

impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanObserverAssignabilityTest.java
Ladicek: Here's a proposal: https://github.com/jakartaee/cdi/pull/744 I went full on the precise mode, as that's just easiest frankly :-)

Ladicek: > their own implementation of `Bean` which may not return `Object` from `getTypes`.

That would be an invalid implementation of `Bean` in my eyes, though the container can compensate for it. All the more reason for using the `BeanConfigurator` API.

Azquelt: > Here's a proposal

That looks good, thank you :-)

I've updated the test to match.

> That would be an invalid implementation of `Bean` in my eyes

Yeah, I agree and it's not the only way you could have an invalid `Bean` implementation either (e.g. returning a non-scope annotation from `getScope()`). I guess we could be more specific about how and when these sorts of errors are handled.

I note that `BeanConfigurator` is still vulnerable to some of the same issues. You could fail to pass `Object` to [`types`][types] as @ljnelson suggested, or pass a non-scope annotation to [`scope`][scope] and we don't currently document what should happen.

[types]: https://jakarta.ee/specifications/cdi/4.0/apidocs/jakarta.cdi/jakarta/enterprise/inject/spi/configurator/beanattributesconfigurator#types(java.lang.reflect.Type...)
[scope]: https://jakarta.ee/specifications/cdi/4.0/apidocs/jakarta.cdi/jakarta/enterprise/inject/spi/configurator/beanattributesconfigurator#scope(java.lang.Class)

impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanObserverAssignabilityTest.java
Azquelt: Now fixed.

impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanObserverAssignabilityTest.java
Ladicek: Good points. Out of scope of this PR for sure, though :-)

manovotn: > If all synthetic beans are beans, and if all beans have the bean type java.lang.Object, then all synthetic beans have the bean type java.lang.Object. That is a straightforward implication in my eyes.

I am of the same opinion.

> The only possible downside is that someone might start asking "why don't we automatically scan for all superclasses and implemented interfaces?", and I wouldn't have a good answer :-)

Because you might want to simulate a scenario where `@Typed` was used and some types might be omitted.

> Yeah, I agree and it's not the only way you could have an invalid Bean implementation either (e.g. returning a non-scope annotation from getScope()). I guess we could be more specific about how and when these sorts of errors are handled.

Well, since you can define arbitrary things if you implement `Bean` yourself, it would be rather extensive to nail it all down...

impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanEventAssignabilityTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+        assertTrue(getCurrentBeanContainer().isMatchingEvent(MyEvent.class, Set.of(NamedLiteral.of("foo")), MyEvent.class, Set.of()),
+                "Event with @Named should match when no qualifiers required");
+
+    }
+    
+    @Test
+    @SpecAssertion(section = BM_BEAN_EVENT_ASSIGNABILITY, id = "ad")
+    public void testEventMatchingDefaultQualifier() {
+        assertTrue(getCurrentBeanContainer().isMatchingEvent(MyEvent.class, Set.of(), MyEvent.class, Set.of(Default.Literal.INSTANCE)),
```
manovotn, line 157: @Azquelt I believe this is incorrect and should be `assertFalse`. With observers, you are *not* adding `@Default` if missing, just `@Any`.

manovotn (changes requested):
Two more notes besides the one code comment below:

- I promised you a Weld branch/pr, which is here -> https://github.com/weld/core/pull/2906. It passes all tests in the PR's current state apart from the one which I believe is wrong.
- I am also pondering whether we should assert something about validity of input such as annotations not being CDI qualifiers or bean types being invalid. On one hand it would be nice, OTOH every invocation of this method will have to perform a solid amount of verification just because we assume users may not know what they are doing :shrug:

Ladicek:
> I am also pondering whether we should assert something about validity of input such as annotations not being CDI qualifiers or bean types being invalid. On one hand it would be nice, OTOH every invocation of this method will have to perform a solid amount of verification just because we assume users may not know what they are doing

We talked about this internally, my personal opinion is a bit different in that it is typical for `BeanManager` methods to validate inputs, so I don't see why these methods shouldn't. Here's my proposal: https://github.com/jakartaee/cdi/pull/748

impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanEventAssignabilityTest.java
manovotn: Oh, you are right, I completely forgot about that. I've tweaked Weld PR accordingly and it now passes all of these tests.

> Events don't get @Default as an extra qualifier ...

Well, the spec bit you posted basically implies they do get `@Default` when no other qualifier is present. Otherwise you wouldn't satisfy the observer resolution condition of OM having a subset of qualifiers that the event has.

manovotn:
> > I am also pondering whether we should assert something about validity of input such as annotations not being CDI qualifiers or bean types being invalid. On one hand it would be nice, OTOH every invocation of this method will have to perform a solid amount of verification just because we assume users may not know what they are doing
> 
> We talked about this internally, my personal opinion is a bit different in that it is typical for `BeanManager` methods to validate inputs, so I don't see why these methods shouldn't. Here's my proposal: [jakartaee/cdi#748](https://github.com/jakartaee/cdi/pull/748)

Yea, added my approval there but @Azquelt might want to take a look as well.


We'll also want some TCK coverage for that - the qualifier validation should be an easy one. 
The unresolved type variable might be tricky as the only other case in TCK has this marked as `testable=false` (see [here](https://github.com/jakartaee/cdi-tck/blob/master/impl/src/main/resources/tck-audit-cdi.xml#L5244)).
Finally legal bean types - that might be harder because we cannot check how was the set of types trimmed; instead we'd have to check that the type was removed and cannot be used to satisfy given required type.

@Azquelt do you wish to look into these or should me or Ladislav take that?

impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanEventAssignabilityTest.java
Azquelt: Yeah, I _think_ the two end up equivalent, which makes it all the more strange that they're specified differently.

manovotn: I agree. Either way, marking this as resolved.
Thanks again for pointing it out

Azquelt:
@manovotn I'm happy to look at it, I'll add it to this PR unless you'd rather merge this part first.

I understood "legal bean types" to refer to just [2.1.2.1][legal-bean-types], which I think is only concerned with the type itself and not whether it's been discovered, trimmed or vetoed?

I'm not entirely sure about what an unresolved type variable is (I have also commented on jakartaee/cdi#748), so I'll struggle to test that unless someone can provide an example of something that wouldn't be a valid event object under that rule.

[legal-bean-types]: https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#legal_bean_types

impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanEventAssignabilityTest.java
Ladicek: We should probably mention this in the javadoc of `isMatchingEvent()` as well. I'll add something to https://github.com/jakartaee/cdi/pull/748

Ladicek:
> I'm not entirely sure about what an unresolved type variable is (I have also commented on https://github.com/jakartaee/cdi/pull/748), so I'll struggle to test that unless someone can provide an example of something that wouldn't be a valid event object under that rule.

The audit file for existing tests says that this provision (or rather, the existing provision on which this one is modeled) is untestable. Maybe there's a way to test it, but I'd be fine with leaving that one uncovered.

manovotn:
> > I'm not entirely sure about what an unresolved type variable is (I have also commented on https://github.com/jakartaee/cdi/pull/748), so I'll struggle to test that unless someone can provide an example of something that wouldn't be a valid event object under that rule.
> 
> The audit file for existing tests says that this provision (or rather, the existing provision on which this one is modeled) is untestable. Maybe there's a way to test it, but I'd be fine with leaving that one uncovered.

Yes that's exactly why I linked the tck file. I would also be fine leaving this untested.

manovotn:
> @manovotn I'm happy to look at it, I'll add it to this PR unless you'd rather merge this part first.
> 

Feel free to add it here. I would rather merge once we release cdi api anyway.

> I understood "legal bean types" to refer to just [2.1.2.1][legal-bean-types], which I think is only concerned with the type itself and not whether it's been discovered, trimmed or vetoed?
>
> [legal-bean-types]: https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#legal_bean_types

Right, it's just a case where a user can pass in a set of types where some would normally be filtered as illegal for given bean

manovotn:
> I've added tests for the changes in [jakartaee/cdi#748](https://github.com/jakartaee/cdi/pull/748).
> 
> The only one that didn't pass against @manovotn's branch was [this one](https://github.com/jakartaee/cdi-tck/pull/526/files#diff-bc73ce456de5cac4811d8bdf91f0420a8b1bed84086af8dd99870eedd32ad6baR230-R240) for having a type variable in the event type. I'd tried to test this in a similar way to [FireEventTest.testTypeVariableEventTypeFails](https://github.com/jakartaee/cdi-tck/blob/046e78eaf5921afd8a51792208c5e30fda444269/impl/src/main/java/org/jboss/cdi/tck/tests/event/fires/FireEventTest.java#L80-L84).

Thanks, I'll check on it.

impl/src/main/java/org/jboss/cdi/tck/tests/beanContainer/assignability/BeanEventAssignabilityTest.java
```diff
@@ -0,0 +1,252 @@
...
+
+    @Test
+    @SpecAssertion(section = BM_BEAN_EVENT_ASSIGNABILITY, id = "bd")
+    public <X> void testEventMatchingTypeVarException() {
+        TypeLiteral<List<X>> varEventType = new TypeLiteral<>() {};
+
+        assertThrows("Type variable in event type", IllegalArgumentException.class,
+                () -> getCurrentBeanContainer().isMatchingEvent(varEventType.getType(), Set.of(), MyBean.class, Set.of()));
+
+        assertThrows("Type variable in observed event type", IllegalArgumentException.class,
```
manovotn, line 238: @Azquelt I might be missing something obvious but which part of specification does this test? The linked section is only for the event (payload) type. I don't see a similar restriction for the observed type?

Azquelt: You're quite right, that's a mistake. I've removed that check.

Ladicek (approved):

Ladicek:
My implementation in ArC also passes.

We should probably merge https://github.com/jakartaee/cdi/pull/748 first and then update the audit files here, but I wouldn't really mind otherwise. Except we'll need a new CDI API release (Alpha2?) first, anyway.

Azquelt:
I've rebased and updated the audit files based on jakartaee/cdi#748 now that it's merged.

manovotn:
@Azquelt `master` branch of CDI TCK repo now has CDI API `4.1.0.Beta1` - please rebase this PR and we should be able to get it merged

Azquelt:
Rebased, squashed and copyright headers updated.

PR #527, opened 2024-01-16
remove the JBoss tagline from license headers

Ladicek:


Ladicek:
For the record, this is how I created this PR:

```bash
find . -name *.java -print0 | xargs -0 sed -i -e '/Home of Professional Open Source/d'
```

There were 3 files that got wiped completely and I had to fix those manually.

Ladicek:
Thanks for listing those files, I agree we should just convert them to LF line endings. Added a 2nd commit that does that.

Azquelt (approved):
Thanks

manovotn (approved):
I've done a quick scan of the changes and it looks good.
The second commit makes sense as well, thanks for noticing that.

PR #532, opened 2024-01-31
Initial update to remove the web integration tests from the core tck

### LINKED ISSUE #529: Remove the web module for the next release
We need to remove the web module from the project for the next release in preparation for moving these tests to a web profile platform tck. Any relevant documentation references need to be updated as well.

- [x] Migrate existing web tests to the platform-tck repo
- [ ] Update the TCK documents to reflect the move \
(https://github.com/jakartaee/cdi-tck/pull/532)
- [ ] Release a Weld implementation
- [ ] Update the sample runner

starksm64:
Fixes #529


dist-build/README.adoc
```diff
@@ -85,11 +83,11 @@ Then:
 
 * `mvn install -f $CDI_TCK_ROOT/artifacts/pom.xml`
 * `mvn -Dtck package -f $CDI_TCK_ROOT/weld/jboss-as/pom.xml` - to enable CDI Extension
-* `mvn clean verify -Dincontainer -Dcdi.tck-4-0.version=${TCK_VERSION} -f $CDI_TCK_ROOT/weld/lang-model-tck-runner/pom.xml`` - to run the TCK against WildFly
+* `mvn clean verify -Dincontainer -Dcdi.tck.version=${TCK_VERSION} -f $CDI_TCK_ROOT/weld/lang-model-tck-runner/pom.xml`` - to run the TCK against WildFly
```
Ladicek, line 86: The `cdi.tck-4.0.version` property is for Weld, right? At the moment, Weld uses `cdi.tck-4-1.version` (and I don't see a PR in https://github.com/weld/core to change it), so this change (here and elsewhere) is wrong.

starksm64: No, there is a separate weld.version for that. How this cdi.tck-4-0.version property was used was to allow one to select a different version of the TCK when running the sample test runner poms. Really this dates back to when one could do minor releases of the TCK without doing a full release, so I just dropped it as it is confusing as to what the purpose is and it just complicates the build and verifying it.

manovotn: I think Ladislav is right here. This command is for Weld to execute TCKs against certain TCK version and the property is defined in Weld [here](https://github.com/weld/core/blob/master/pom.xml#L63) as `cdi.tck-4-1.version`

The reason it's so explicit is that in some point in time Weld had to run against two different TCK majors (around CDI 2/3 IIRC). That being said, I wouldn't mind changing it to `cdi.tck.version` if that's what you'd prefer.

starksm64: Ok, that is confusing. Let me look at exactly Weld uses it and how this fits into how the TCK is released now.

starksm64: I can't see why we would want to take the CDI TCK version used by the sample TCK runner from the weld-core-parent pom cdi.tck-4-1.version value. What is missing in the current changes is a default value for the cdi.tck.version and not using filtering on this property when the jboss-runner.pom is copied to the dist weld/jboss-tck-runner/pom.xml file so that one can override the CDI TCK version used. We can talk about it in today's call.



```
  <properties>
    <!-- This matches the htmlunit version in TCK -->
    <htmlunit.version>2.50.0</htmlunit.version>
    <cdi.tck.version>${project.version}</cdi.tck.version>
  </properties>
```

manovotn (approved):

PR #609, opened 2024-10-09
Add TCK coverage for unproxyability of sealed classes

### LINKED ISSUE #607: Create TCK coverage for unproxiability of sealed classes
In 4.1 we clarified that sealed types and interfaces are not proxyable ([spec link](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#unproxyable)).
However, we added no TCK coverage at the time as baseline JDK was 11 and I don't see any tracking issue for that.

Since that's no longer the case, we should add a few tests for the upcoming version.

manovotn:
Fixes #607

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/lookup/clientProxy/unproxyable/sealed/InjectingBean1.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/lookup/clientProxy/unproxyable/sealed/InjectingBean1.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2024, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/lookup/clientProxy/unproxyable/sealed/MyNormalScopedBean.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.cdi.tck.tests.lookup.clientProxy.unproxyable.sealed;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public sealed class MyNormalScopedBean permits MyNormalScopedSubclass {
```
Ladicek, line 22: Should be called `MyNormalScoped` for consistency with other classes in this test.

manovotn: Done

PR #623, opened 2025-02-27
add Lang Model TCK coverage for sealed classes

Ladicek:
TCK coverage for https://github.com/jakartaee/cdi/pull/857

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,7 +96,7 @@
 
     <properties>
         <!-- CDI API -->
-        <cdi.api.version>5.0.0.Alpha1-SNAPSHOT</cdi.api.version>
+        <cdi.api.version>5.0.0.Alpha1</cdi.api.version>
         <maven.compiler.release>17</maven.compiler.release>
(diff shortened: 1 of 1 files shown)


Ladicek:
CI is supposed to fail ATM, because of a `SNAPSHOT` dependency.

manovotn:
> CI is supposed to fail ATM, because of a `SNAPSHOT` dependency.

@Ladicek @Azquelt unless you have something against it, I'd propose we start doing `Alpha` releases (for both, CDI and TCK) like we did in previous versions. That will allow us to merge TCK PRs and impls can also consume it to keep up to date with spec changes.

Ladicek:
Agree.

Azquelt:
I agree as well

manovotn:
After sorting out some issues with SSH key in Jenkins, the CDI API Alpha1 is now in Central.
I've added a commit updating it - the CI should now turn green.

manovotn (approved):

Issue #259, opened 2021-08-24
Look into updating the audit files


starksm64:
The TCK has various tck-audit-*.xml files that conform to the http://jboss.com/products/weld/tck/audit schema that providing a mapping from specification assertions and associated text to the actual unit test. This has gotten out of date, and it should be a goal to bring these files back into alignment with the specifications.

With the move to asciidoc, it may be possible to create a converter that allows for the generation of the audit files.

manovotn:
Thanks for filing the issue. Had it on my TODO but went offline earlier yesterday.

To fill in some details:
* TCKs  audit files were for CDI spec, interceptors and managed bean definition although we are mainly looking at having it working for CDI
  * They can be seen [here](https://github.com/eclipse-ee4j/cdi-tck/tree/master/impl/src/main/resources)
* Upon building TCKs, a file is generated locally with a name of `Sections.java` that links the audit XML file with annotation values used on actual tests
* For TCK 1.2 and 2.0 we specifically generated an HTML version of spec with links to TCK. An example of such doc can be seen [here](https://docs.jboss.org/cdi/spec/2.0/cdi-spec-with-assertions.html#default_bean_discovery)
  * So you could either browse this doc or you could just open up the audit file and search there and then `grep` TCKs for needed assertions

The separate project that is used to generate the links between tests and audit file is this one - https://github.com/jboss/jboss-test-audit. To my knowledge, it was only ever used by CDI and bean validation specs but it shouldn't be an issue to export it under jakarta repos if we need/want to.

Personally, big +1 for putting this back on track. This makes working with TCKs and looking for certain tests much easier. Although it will be quite some effort due to ongoing re-shuffling of spec due to Lite addition.

Ladicek:
I was able to build a proof of concept. Here's how the outcome looks like: https://ladicek.github.io/cdi-tck/#legal_bean_types

The proof of concept code itself is here:l https://github.com/Ladicek/specref

The required changes to CDI spec are here: https://github.com/Ladicek/cdi-spec/commits/specref

And the required changes to CDI TCK are here: https://github.com/Ladicek/cdi-tck/commits/specref

starksm64:
@Ladicek So you also created a io.smallrye.specref-core repo as well, because I'm not finding that in the io.smallrye org. Is that an asciidoc extension or converter?

Ladicek:
[...]

PR #644, opened 2025-10-02
Add Java 25 to CI

manovotn:


manovotn:
> > I have no idea why this failure occurs:
> > ```
> > Error:  Failed to execute goal org.apache.maven.plugins:maven-install-plugin:3.1.2:install (default-install) on project cdi-tck-core-impl: Failed to install artifact jakarta.enterprise:cdi-tck-core-impl:html:coverage-cdi:5.0.0-SNAPSHOT: /home/runner/work/cdi-tck/cdi-tck/target/coverage-cdi.html -> [Help 1]
> > ```
> 
> It appears the files are not being generated. The `build-helper-maven-plugin` might be triggering too late or not at all? Although I don't know why either. I am trying to figure it out...

Ok, I have no idea what's wrong either :shrug: 

The build-helper plugin executes just fine but the files aren't present.
The issue seems to be that during `compile` the files don't get created at all; the test audit just don't happen.

In earlier JDKs, during compiler plugin execution you would see an output such as this:

```
Outputting to /home/manovotn/GitRepo/cdi-tck/impl/../target/
Reading spec assertions from audit file/s: /home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-cdi.xml,/home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-mb.xml,/home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-int.xml
Reading spec assertions from audit file/s: /home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-cdi.xml,/home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-mb.xml,/home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-int.xml
Generate coverage report: Jakarta EE: Managed Beans
Reading spec assertions from audit file/s: /home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-cdi.xml,/home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-mb.xml,/home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-int.xml
Generate coverage report: Jakarta Contexts and Dependency Injection 3.0
Reading spec assertions from audit file/s: /home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-cdi.xml,/home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-mb.xml,/home/manovotn/GitRepo/cdi-tck/impl/src/main/resources/tck-audit-int.xml
Generate coverage report: Jakarta Interceptors Specification

```
In JDK 25, it's not there at all.
I am not seeing any error though, not even when running maven with debug...

manovotn:
Ok, so the annotation processor (`CoverageProcessor`) is not invoked at all with JDK 25 for reasons beyond my understanding.
Explicitly adding it as a compiler argument fixes the issue though.

We should probably consider if we want to keep the audit in place - the [dependency](https://github.com/jboss/jboss-test-audit) it uses is old and archived and we also have a long standing [issue](https://github.com/jakartaee/cdi-tck/issues/259) for update of this approach that has not been worked on.

PR #659, opened 2025-11-07
add TCK tests for eager initialization (`@Eager`)

Ladicek:
TCK tests for https://github.com/jakartaee/cdi/pull/911

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/eager/bean/EagerBeanTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/eager/bean/EagerBeanTest.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.cdi.tck.tests.eager.bean;
+
+import static org.jboss.cdi.tck.cdi.Sections.BEAN;
+import static org.jboss.cdi.tck.cdi.Sections.EAGER_INITIALIZATION;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.cdi.tck.AbstractTest;
+import org.jboss.cdi.tck.shrinkwrap.WebArchiveBuilder;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.jboss.test.audit.annotations.SpecAssertion;
+import org.jboss.test.audit.annotations.SpecVersion;
+import org.testng.annotations.Test;
+
+@SpecVersion(spec = "cdi", version = "5.0")
+public class EagerBeanTest extends AbstractTest {
+    @Deployment
+    public static WebArchive createTestArchive() {
+        return new WebArchiveBuilder()
+                .withTestClassPackage(EagerBeanTest.class)
+                .build();
+    }
+
+    @Test
+    @SpecAssertion(section = EAGER_INITIALIZATION, id = "aa")
+    @SpecAssertion(section = EAGER_INITIALIZATION, id = "b")
+    @SpecAssertion(section = EAGER_INITIALIZATION, id = "c")
+    public void testUnqualified() {
+        assertTrue(EagerBean.constructed);
+
+        assertFalse(LazyBean.constructed);
+        LazyBean reference = getContextualReference(LazyBean.class);
+        assertNotNull(reference);
+        assertFalse(LazyBean.constructed);
+        assertNotNull(reference.toString());
+        assertTrue(LazyBean.constructed);
+    }
+
+    @Test
+    @SpecAssertion(section = EAGER_INITIALIZATION, id = "aa")
+    @SpecAssertion(section = EAGER_INITIALIZATION, id = "b")
+    @SpecAssertion(section = EAGER_INITIALIZATION, id = "c")
+    public void testQualified() {
+        assertTrue(QualifiedEagerBean.constructed);
+
+        assertFalse(QualifiedLazyBean.constructed);
+        QualifiedLazyBean reference = getContextualReference(QualifiedLazyBean.class, AQualifier.Literal.INSTANCE);
+        assertNotNull(reference);
+        assertFalse(QualifiedLazyBean.constructed);
+        assertNotNull(reference.toString());
+        assertTrue(QualifiedLazyBean.constructed);
+    }
+
+    @Test
+    @SpecAssertion(section = BEAN, id = "bf")
+    public void testMetadata() {
+        assertTrue(getUniqueBean(EagerBean.class).isEager());
+        assertTrue(getUniqueBean(QualifiedEagerBean.class, AQualifier.Literal.INSTANCE).isEager());
+
+        assertFalse(getUniqueBean(LazyBean.class).isEager());
+        assertFalse(getUniqueBean(QualifiedLazyBean.class, AQualifier.Literal.INSTANCE).isEager());
+    }
+}
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/eager/stereotype/broken/dependentAndEagerOnStereotype/DependentEagerBean.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/eager/stereotype/broken/dependentAndEagerOnStereotype/DependentEagerBean.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.cdi.tck.tests.eager.stereotype.broken.dependentAndEagerOnStereotype;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+@EagerDependentStereotype
+public class DependentEagerBean {
+}
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/eager/synthetic/bean/SyntheticBeanExtension.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/eager/synthetic/bean/SyntheticBeanExtension.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.cdi.tck.tests.eager.synthetic.bean;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.build.compatible.spi.BuildCompatibleExtension;
+import jakarta.enterprise.inject.build.compatible.spi.Synthesis;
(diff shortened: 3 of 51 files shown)


Ladicek:
This is a draft, because:

- the spec PR (https://github.com/jakartaee/cdi/pull/911) wasn't merged yet
- there's a SNAPSHOT dependency for now

Other than that, this should be complete I believe.

Ladicek:
Rebased. No SNAPSHOT dependency anymore.

impl/src/main/java/org/jboss/cdi/tck/tests/reserve/basic/ReserveMetadataTest.java
```diff
@@ -0,0 +1,39 @@
...
+import static org.testng.Assert.assertTrue;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.cdi.tck.AbstractTest;
+import org.jboss.cdi.tck.shrinkwrap.WebArchiveBuilder;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.jboss.test.audit.annotations.SpecAssertion;
+import org.testng.annotations.Test;
+
+public class ReserveMetadataTest extends AbstractTest {
```
manovotn, line 25: Missing `@SpecVersion`

impl/src/main/java/org/jboss/cdi/tck/tests/eager/synthetic/bean/SyntheticBeanExtension.java
```diff
@@ -0,0 +1,45 @@
...
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.build.compatible.spi.BuildCompatibleExtension;
+import jakarta.enterprise.inject.build.compatible.spi.Synthesis;
+import jakarta.enterprise.inject.build.compatible.spi.SyntheticComponents;
+import jakarta.enterprise.inject.build.compatible.spi.Types;
+
+public class SyntheticBeanExtension implements BuildCompatibleExtension {
+    @Synthesis
+    public void synthesize(SyntheticComponents syn, Types types) {
```
manovotn: Nitpicking but `Types` param can be removed

impl/src/main/java/org/jboss/cdi/tck/tests/reserve/selection/stereotype/StereotypeWithReserveSelectedByPriorityTest.java
```diff
@@ -46,4 +53,17 @@ public void testStereotypeReserveIsEnabled() {
...
+
+    @Test
+    @SpecAssertion(section = BEAN, id = "be")
+    public void testMetadata() {
+        assertTrue(resolve(SomeInterface.class).isReserve());
+        assertTrue(resolve(SomeInterface.class, ProducedByMethod.Literal.INSTANCE).isReserve());
+        assertTrue(resolve(SomeInterface.class, ProducedByField.Literal.INSTANCE).isReserve());
+    }
+
+    private Bean<?> resolve(Class<?> beanType, Annotation... qualifiers) {
```
manovotn: There should be a `getUniqueBean` method coming from the `AbstractTest` that does just this.

manovotn (commented):
Just a few small notes. Otherwise looks good

impl/src/main/java/org/jboss/cdi/tck/tests/reserve/selection/stereotype/StereotypeWithReserveSelectedByPriorityTest.java
Ladicek: OMG :facepalm: Thanks!

impl/src/main/java/org/jboss/cdi/tck/tests/reserve/basic/ReserveMetadataTest.java
Ladicek: Done.

impl/src/main/java/org/jboss/cdi/tck/tests/eager/synthetic/bean/SyntheticBeanExtension.java
Ladicek: Done.

impl/src/main/java/org/jboss/cdi/tck/tests/reserve/selection/stereotype/StereotypeWithReserveSelectedByPriorityTest.java
Ladicek: Done.

manovotn (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/eager/bean/EagerBeanTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    @SpecAssertion(section = EAGER_INITIALIZATION, id = "aa")
+    @SpecAssertion(section = EAGER_INITIALIZATION, id = "b")
+    @SpecAssertion(section = EAGER_INITIALIZATION, id = "c")
+    public void testUnqualified() {
+        assertTrue(EagerBean.constructed);
+
+        assertFalse(LazyBean.constructed);
```
Azquelt, line 42: Is this behavior (that a non-eager bean is not constructed until it is needed) a spec requirement?

impl/src/main/java/org/jboss/cdi/tck/tests/eager/stereotype/broken/dependentAndEagerOnStereotype/DependentEagerBean.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.cdi.tck.tests.eager.stereotype.broken.dependentAndEagerOnStereotype;
+
+import jakarta.enterprise.context.Dependent;
+
+@Dependent
+@EagerDependentStereotype
+public class DependentEagerBean {
```
Azquelt: This bean is `@Eager` and `@Dependent`, which is a different problem than we're trying to test here. I suggest changing this to `ApplicationScoped`.

The stereotype having `@Eager` and `@Dependent` together should be enough to cause the definition error, even if there are no eager beans with non-permitted scopes.

impl/src/main/java/org/jboss/cdi/tck/tests/eager/synthetic/bean/SyntheticBeanExtension.java
```diff
@@ -0,0 +1,44 @@
...
+                .scope(ApplicationScoped.class)
+                .qualifier(AQualifier.class)
+                .eager(true)
+                .createWith(QualifiedEagerClassCreator.class);
+
+        syn.addBean(QualifiedLazyClass.class)
+                .type(QualifiedLazyClass.class)
+                .scope(ApplicationScoped.class)
+                .qualifier(AQualifier.class)
+                .createWith(QualifiedLazyClassCreator.class);
```
Azquelt, line 43: Maybe add `eager(false)` to one of the lazy beans to ensure the argument is not ignored.

impl/src/main/java/org/jboss/cdi/tck/tests/eager/bean/EagerBeanTest.java
Ladicek: Not a directly stated one, no, but it feels implied. A bean is only instantiated through `Bean.create()` (actually `Contextual.create()`), and if noone calls that directly or indirectly, no instance exists.

impl/src/main/java/org/jboss/cdi/tck/tests/eager/stereotype/broken/dependentAndEagerOnStereotype/DependentEagerBean.java
Ladicek: If the bean was `@ApplicationScoped` directly, the scope on the stereotype would be ignored (see [2.8.1.1. Declaring the default scope for a stereotype](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#stereotype_default_scope)). So that wouldn't work.

But I'm not sure why the bean has a scope annotation. A stereotype is a bean defining annotation, so that should be enough... I think we should just remove `@Dependent` here.

impl/src/main/java/org/jboss/cdi/tck/tests/eager/synthetic/bean/SyntheticBeanExtension.java
Ladicek: That makes sense, just as extra test coverage. Will do.

impl/src/main/java/org/jboss/cdi/tck/tests/eager/stereotype/broken/dependentAndEagerOnStereotype/DependentEagerBean.java
Ladicek: Ah I see what you mean now. This test verifies that

> If a stereotype declares both `@Eager` and a default scope for which eager initialization is not permitted, the container automatically detects the problem and treats it as a definition error.

And the bean scope here doesn't really matter. So we can make it `@ApplicationScoped` and rely on the container to validate the stereotype correctly. Good point. I'll change the bean to `@ApplicationScoped`.

Ladicek: Done.

impl/src/main/java/org/jboss/cdi/tck/tests/eager/synthetic/bean/SyntheticBeanExtension.java
Ladicek: Done.

impl/src/main/java/org/jboss/cdi/tck/tests/eager/bean/EagerBeanTest.java
Azquelt: Ok. I'm a _little_ concerned that we're relying on something that isn't an explicit spec requirement but I think it's a reasonable assumption and I guess in the worst case we get a challenge for it.

Azquelt (approved):

Ladicek:
@manovotn See the last comments from @Azquelt -- I hope the one change of scope is OK for you, but I vaguely recall discovery works differently in Weld, so would be good to have another look.

impl/src/main/java/org/jboss/cdi/tck/tests/eager/bean/EagerBeanTest.java
manovotn: > Not a directly stated one, no, but it feels implied.

It is not stated anywhere and theoretically, you could have a completely eager-init implementation.
However, such impl would be very slow in terms of performance.

Besides, we already assume the initialization is lazy - which is why we introduce `@Eager` in the first place :shrug: 
Perhaps we should consider stating that in the specification as well...though I didn't think that through further. It could have other implications as well.

manovotn:
> @manovotn See the last comments from @Azquelt -- I hope the one change of scope is OK for you, but I vaguely recall discovery works differently in Weld, so would be good to have another look.

Yea, that breaks my current implementation. Let me take a closer look at what's going on there...

manovotn:
> > @manovotn See the last comments from @Azquelt -- I hope the one change of scope is OK for you, but I vaguely recall discovery works differently in Weld, so would be good to have another look.
> 
> Yea, that breaks my current implementation. Let me take a closer look at what's going on there...

@Ladicek the discovery is not the issue here, it's the validation.
IIRC, Weld does validation per-bean, not on the stereotype itself. And the bean it perfectly valid.

I think we had a similar discussion regarding alternative and reserve on a stereotype and I can see I added

manovotn:
> > > @manovotn See the last comments from @Azquelt -- I hope the one change of scope is OK for you, but I vaguely recall discovery works differently in Weld, so would be good to have another look.
> > 
> > 
> > Yea, that breaks my current implementation. Let me take a closer look at what's going on there...
> 
> @Ladicek the discovery is not the issue here, it's the validation. IIRC, Weld does validation per-bean, not on the stereotype itself. And the bean it perfectly valid.
> 
> I think we had a similar discussion regarding alternative and reserve on a stereotype and I can see I added

OK, fixed now, thanks for pointing it out :+1:

PR #676, opened 2026-01-26
improve annotations and types assertions

Ladicek:
All tests now use assertion methods from `org.jboss.cdi.tck.Assert`. The helper methods on `AbstractTest`, which were in all but 1 cases used just as an argument to `assertTrue()`, are removed.

There was one bug in `AbstractTest.annotationSetMatches()` that is fixed now: the method asserted correct number of elements, but didn't assert correct elements themselves. This bugfix may reveal issues in implementations, but ArC and Weld are correct. Further, the `ProducerMethodWithDefaultNameTest.testProducerMethodQualifiers()` test was wrong (fortunately, all other tests were fine): it asserted that the producer method bean has qualifiers `Any` and `Default`, while it should have `Any` and `Funny`. This bug is fixed as well.

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/AbstractTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/AbstractTest.java
@@ -15,13 +15,7 @@ package org.jboss.cdi.tck;
 
 import java.io.IOException;
 import java.lang.annotation.Annotation;
-import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 
 import jakarta.enterprise.context.spi.Context;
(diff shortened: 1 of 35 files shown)


manovotn (approved):

Azquelt (approved):

manovotn:
@Ladicek @Azquelt I have just realized (or rather ran headlong into on Weld) that the Platform TCK EE tests (jakartaee/platform-tck) still use the old signatures, causing `NoSuchMethodError` at runtime.

The fix needs to happen in the Platform TCK repo, but since we can't know when they'll release an EE 12 compatible version, we might want to keep the old method variants in the CDI TCK as deprecated bridges (marked for removal), so both old and new callers work. WDYT? We could create an issue tracking their removal as soon as we know there is any platform TCK release in the wild.

Obviously I can work around this in Weld by either leaving out the platform tests until there is a release or attempting to run these in a separate profile but that is clumsy as I will need TCK Alpha2 to run platform TCK and then whichever is the latest TCK to run our own tests :shrug:

PR #678, opened 2026-01-30
 TCK tests for improved specification of `InjectionPoint`

Ladicek:
TCK counterpart of https://github.com/jakartaee/cdi/pull/931

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/lookup/injectionpoint/dynamic/notinjected/DynamicInjectionPointNotInjectedTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/lookup/injectionpoint/dynamic/notinjected/DynamicInjectionPointNotInjectedTest.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.cdi.tck.tests.lookup.injectionpoint.dynamic.notinjected;
+
+import static org.jboss.cdi.tck.cdi.Sections.INJECTION_POINT;
+import static org.jboss.cdi.tck.util.Assert.assertAnnotationsMatch;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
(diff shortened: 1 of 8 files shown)


Azquelt (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/lookup/injectionpoint/dynamic/notinjected/DynamicInjectionPointNotInjectedTest.java
```diff
@@ -0,0 +1,98 @@
...
+        assertEquals(bar.getFoo().getInjectionPoint().getType(), Foo.class);
+        assertEquals(bar.getNiceFooByType().getInjectionPoint().getType(), NiceFoo.class);
+        assertEquals(bar.getNiceFooByQualifier().getInjectionPoint().getType(), Foo.class);
+        assertEquals(bar.getNiceFooByAllQualifiers().getInjectionPoint().getType(), Foo.class);
+    }
+
+    @Test
+    @SpecAssertion(section = INJECTION_POINT, id = "bca")
+    public void testInjectionPointGetQualifiers() {
+        assertAnnotationsMatch(bar.getFoo().getInjectionPoint().getQualifiers(), Default.Literal.INSTANCE);
```
manovotn, line 52: I found out Weld doesn't actually do this.
We'd return an empty set as no qualifier was actually *required* here.

Now, the fix is trivial but I wonder if we really want to return `@Default`?
In practice, doing so blurs the distinction between user explicitly requiring `@Default` and requiring nothing.

Ladicek: ArC didn't do it either, but it is a bug in my eyes. There's no distinction between users explicitly requiring `@Default` and requiring nothing -- if an injection point has no qualifier, it has `@Default`.

Ladicek: See also https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#injection_point_default_qualifier

manovotn: Yea, I think I agree :)

manovotn (approved):

PR #679, opened 2026-02-04
add TCK test for specifying parameterized types in @Registration

Ladicek:
TCK for https://github.com/jakartaee/cdi/pull/937

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/registration/RegistrationExtension.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/registration/RegistrationExtension.java
@@ -13,6 +13,8 @@
  */
 package org.jboss.cdi.tck.tests.build.compatible.extensions.registration;
 
+import java.util.Collection;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import jakarta.enterprise.inject.build.compatible.spi.BeanInfo;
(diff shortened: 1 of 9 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/registration/RegistrationExtension.java
```diff
@@ -38,13 +43,45 @@ public void beans(BeanInfo bean) {
...
+    @Registration(types = MyGenericServiceOfT.class)
+    public void genericBeans2(BeanInfo bean) {
+        genericBeanCounter.incrementAndGet();
+    }
+
+    static class MyGenericServiceOfString extends TypeLiteral<MyGenericService<String>> {
+    }
+
+    static class MyGenericServiceOfT<T> extends TypeLiteral<MyGenericService<T>> {
+    }
```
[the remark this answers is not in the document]
Ladicek: OK, let's take a step back. `@Registration` is relatively different for beans and for observers. You don't speak about observers, so let's focus on beans:

> In the following text, the term _expected types_ denotes the set of types defined by the `types` member of the `@Registration` annotation.
>
> ...
>
> If a `@Registration` method has a parameter of type `BeanInfo`, the method is called once for each bean whose set of bean types contains at least one _expected type_.

So if I have a bean with a bean type `List<String>`, it does _not_ have a bean type of `List<T>` where `T` is a type variable. Therefore, this

> To me it seems odd that `MyGenericServiceOfT` matches `MyGenericService<T>` but not `MyGenericService<String>`.

is an incorrect expectation, just per the existing specification rules.

The rules are definitely _not_ the same (or even close to) the observer assignability rules. If we wanted to change them, that would be an entirely different discussion.

manovotn (approved):

Ladicek:
Rebased and squashed the commits.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/registration/RegistrationExtension.java
```diff
@@ -67,6 +100,10 @@ public void test(Messages msg) {
             msg.error("Should see 1 observer declared in class that implements MyService");
         }
```
manovotn: @Ladicek I think this is actually wrong.
You added more observers to `MyServiceFoo` so this should now be `observerCounter.get() != 4`

Ladicek: Hmm, you're right. I wonder why this didn't fail locally... Thanks, will take a look.

Ladicek: Fixed, thank you!

The issue was that the types here are interface types and my code for looking up supertypes didn't include `Object` in the result if the input was an interface type. Fixed that too.

manovotn: Thanks, I found this while running against my Weld implementation :)

manovotn (approved):

PR #708, opened 2026-04-16
Add TCK tests for SyntheticBeanBuilder.withInjectionPoint()

manovotn:
Fixes #699 

These are created based on a set of tests I was first developing for Weld implementation.

~~For starters, it is a draft as I need to take a second look at it myself but I can't really focus on it anymore for now, so I'll try to do that tomorrow :)~~ (EDIT: done)
I tried to cover all of the various cases I could think of but I could definitely use @Ladicek to take a look if there is any other case you had in mind. 

One thing I am not really comfortable with is the fact that we now want impls to always throw exception if you ask for a type/qualifier combination that wasn't previously stated as required.
It makes sense for build-time approach with optimizations, but while implementing this in Weld, I basically have to put in place restrictions that don't make much sense - there is nothing preventing me from giving user access to the bean he previously didn't state if the bean exists.
To me, the main value lies in early validation of the requested bean in the CDI app but I am not sure I am happy with preventing user from accessing other beans should they wish to :shrug: 
WDYT @Ladicek @Azquelt, wouldn't it make more sense to require the early validation but allow to resolve any bean at your own risk?

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsAnnotationInfo/AnnotationInfoExtension.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsAnnotationInfo/AnnotationInfoExtension.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.cdi.tck.tests.build.compatible.extensions.syntheticBeanInjectionsAnnotationInfo;
+
+import java.lang.annotation.Annotation;
+
+import jakarta.enterprise.context.Dependent;
+import jakarta.enterprise.inject.build.compatible.spi.AnnotationBuilder;
+import jakarta.enterprise.inject.build.compatible.spi.BuildCompatibleExtension;
+import jakarta.enterprise.inject.build.compatible.spi.Synthesis;
+import jakarta.enterprise.inject.build.compatible.spi.SyntheticComponents;
+
+public class AnnotationInfoExtension implements BuildCompatibleExtension {
+    @Synthesis
+    public void synthesize(SyntheticComponents syn) {
+        syn.addBean(SyntheticPojo.class)
+                .type(SyntheticPojo.class)
+                .scope(Dependent.class)
+                .withInjectionPoint(Widget.class, new Annotation[0])
+                .withInjectionPoint(Widget.class, AnnotationBuilder.of(Nice.class).build())
+                .createWith(SyntheticPojoCreator.class);
+    }
+}
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsDependent/SyntheticInjectionsDependentTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsDependent/SyntheticInjectionsDependentTest.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.cdi.tck.tests.build.compatible.extensions.syntheticBeanInjectionsDependent;
+
+import static org.testng.Assert.assertEquals;
+
+import jakarta.enterprise.inject.Instance;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.cdi.tck.AbstractTest;
+import org.jboss.cdi.tck.cdi.Sections;
+import org.jboss.cdi.tck.shrinkwrap.WebArchiveBuilder;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.jboss.test.audit.annotations.SpecAssertion;
+import org.jboss.test.audit.annotations.SpecVersion;
+import org.testng.annotations.Test;
+
+@SpecVersion(spec = "cdi", version = "5.0")
+public class SyntheticInjectionsDependentTest extends AbstractTest {
+    @Deployment
+    public static WebArchive createTestArchive() {
+        return new WebArchiveBuilder()
+                .withTestClassPackage(SyntheticInjectionsDependentTest.class)
+                .withBuildCompatibleExtension(SyntheticInjectionsDependentExtension.class)
+                .build();
+    }
+
+    @Test
+    @SpecAssertion(section = Sections.SYNTHESIS_PHASE, id = "ch", note = "@Dependent beans from SyntheticInjections destroyed with synthetic bean")
+    public void testDependentBeanLifecycle() {
+        TrackedBean.reset();
+        SyntheticPojo.reset();
+
+        Instance<Object> lookup = getCurrentBeanContainer().createInstance();
+
+        assertEquals(TrackedBean.createdCount.get(), 0);
+        assertEquals(TrackedBean.destroyedCount.get(), 0);
+
+        Instance.Handle<SyntheticPojo> handle = lookup.select(SyntheticPojo.class).getHandle();
+        handle.get();
+
+        // TrackedBean was obtained during creation
+        assertEquals(TrackedBean.createdCount.get(), 1);
+        assertEquals(TrackedBean.destroyedCount.get(), 0);
+
+        handle.destroy();
+
+        // After synthetic bean destruction:
+        // - TrackedBean created in creator should be destroyed
+        // - TrackedBean created in disposer is also destroyed (disposer dependents
+        //   are destroyed when disposal completes)
+        assertEquals(TrackedBean.createdCount.get(), 2);
+        assertEquals(TrackedBean.destroyedCount.get(), 2);
+        assertEquals(SyntheticPojo.destroyedCount.get(), 1);
+    }
+}
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsInjectionPoint/SyntheticInjectionsInjectionPointExtension.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsInjectionPoint/SyntheticInjectionsInjectionPointExtension.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.cdi.tck.tests.build.compatible.extensions.syntheticBeanInjectionsInjectionPoint;
+
+import java.lang.annotation.Annotation;
+
+import jakarta.enterprise.context.Dependent;
+import jakarta.enterprise.inject.build.compatible.spi.BuildCompatibleExtension;
+import jakarta.enterprise.inject.build.compatible.spi.Synthesis;
+import jakarta.enterprise.inject.build.compatible.spi.SyntheticComponents;
+import jakarta.enterprise.inject.spi.InjectionPoint;
+
+public class SyntheticInjectionsInjectionPointExtension implements BuildCompatibleExtension {
+    @Synthesis
+    public void synthesize(SyntheticComponents syn) {
+        syn.addBean(SyntheticPojo.class)
+                .type(SyntheticPojo.class)
+                .scope(Dependent.class)
+                .withInjectionPoint(InjectionPoint.class, new Annotation[0])
+                .createWith(SyntheticPojoCreator.class)
+                .disposeWith(SyntheticPojoDisposer.class);
+    }
+}
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsInjectionPoint/SyntheticInjectionsInjectionPointTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsInjectionPoint/SyntheticInjectionsInjectionPointTest.java
@@ -10,8 +10,11 @@
 package org.jboss.cdi.tck.tests.build.compatible.extensions.syntheticBeanInjectionsInjectionPoint;
 
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertNull;
+
+import java.lang.annotation.Annotation;
 
 import jakarta.enterprise.inject.Instance;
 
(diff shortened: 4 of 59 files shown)


manovotn:
The PR is now ready for review.

WRT https://github.com/jakartaee/cdi/pull/964 - the tests here can be easily adapter (just removing the workaround) to account for the new variant. I just kept it as-is to have it build and be able to test it without snapshots.

impl/src/main/resources/tck-audit-cdi.xml
```diff
@@ -6601,6 +6601,42 @@
         <assertion id="b">
             <text>Test that synthesis observers are notified of fired events</text>
         </assertion>
+        <assertion id="ca">
```
manovotn, line 6604: I wasn't sure what to add to the audit file as we don't really cover these parts in the spec but instead do so in the javadoc. I am open to better suggestions :)

Ladicek:
> One thing I am not really comfortable with is the fact that we now want impls to always throw exception if you ask for a type/qualifier combination that wasn't previously stated as required. It makes sense for build-time approach with optimizations, but while implementing this in Weld, I basically have to put in place restrictions that don't make much sense - there is nothing preventing me from giving user access to the bean he previously didn't state if the bean exists. To me, the main value lies in early validation of the requested bean in the CDI app but I am not sure I am happy with preventing user from accessing other beans should they wish to 🤷 WDYT @Ladicek @Azquelt, wouldn't it make more sense to require the early validation but allow to resolve any bean at your own risk?

I didn't review this PR yet, but I want to reply to this: I very much disagree. The design here aligns with producer methods. If you want to lookup arbitrary beans, you have to inject `Instance<Object>` explicitly. It is possible, but you don't get it out of the box, you have to ask for it.

In other words, you're basically pushing back to the `Instance<Object>` approach we had originally. We can just revert the PR that adds `withInjectionPoint()`, if that's what you wish, but you can't have it both ways :-)

manovotn:
> In other words, you're basically pushing back to the `Instance<Object>` approach we had originally. We can just revert the PR that adds `withInjectionPoint()`, if that's what you wish, but you can't have it both ways :-)

Not really @Ladicek

We are deprecating the previous approach so I assume that at some point in the future, we will no longer have that method whatsoever. And that's OK, I take into consideration that we only retain this approach.

More importantly, I understand the value build-time approach gets from the *validation* of those injection points.
In Quarkus, that is what allows us to take note of what beans are required and cannot be optimized away.

However, do you actually get any value from _enforcing_ runtime failure on a bean that is present anyway? Genuine question because I don't think you do. And BTW this is not just a custom `Foo` bean that you added elsewhere but also built-in beans like `BeanContainer`. Perhaps I am just missing something though :)

Ladicek:
Well, if you honestly think it's good for the specification to allow different behaviors _of such a basic thing as dependency injection_ in different implementations...

manovotn:
> Well, if you honestly think it's good for the specification to allow different behaviors _of such a basic thing as dependency injection_ in different implementations...

Not sure I follow? That's not what I said.

Ladicek:
This is exactly what you're suggesting in

> I am not sure I am happy with preventing user from accessing other beans should they wish to

Azquelt:
> One thing I am not really comfortable with is the fact that we now want impls to always throw exception if you ask for a type/qualifier combination that wasn't previously stated as required. It makes sense for build-time approach with optimizations, but while implementing this in Weld, I basically have to put in place restrictions that don't make much sense - there is nothing preventing me from giving user access to the bean he previously didn't state if the bean exists.

I would argue that the main value is consistency and portability. It makes it harder to write an extension that e.g. works on Weld but doesn't work on Quarkus.

manovotn:
> I would argue that the main value is consistency and portability. It makes it harder to write an extension that e.g. works on Weld but doesn't work on Quarkus.

But that's not portability between Lite and Full, just between CDI and Quarkus where Quarkus' optimization contradicts other spec requirements :shrug:

> This is exactly what you're suggesting in
> 
> > I am not sure I am happy with preventing user from accessing other beans should they wish to

Still nope. You can already access any bean you want to via `Instance` all over the place.
Adding a hard check here won't change that.

I guess I don't like that the two APIs (PE and BCE) are just diverging in their capabilities here - PE allows you to do the validation as well (you can add IPS which are used just for validation which is exactly like `withInjectionPoints`) but you can then resolve any bean while BCE is slapping your fingers eagerly.

Of course I understand that without this, the requirement to declare IPs isn't enforceable - I was just questioning the usability of the API from any implementation that does not optimize away unused beans.
And assuming we didn't enforce the failure, impls that want to have this optimization could still attempt the resolution and if the bean isn't present, they could blow up with precise exception detailing on what went wrong.

That said, I won't insist on this and seeing as others are OK with this API, we can keep it as is - it's just that I wanted to share my experience when I was writing these TCKs :)

Ladicek:
> > I would argue that the main value is consistency and portability. It makes it harder to write an extension that e.g. works on Weld but doesn't work on Quarkus.
> 
> But that's not portability between Lite and Full, just between CDI and Quarkus where Quarkus' optimization contradicts other spec requirements 🤷

No it doesn't. We currently have what could arguably be described as a bug in Quarkus where we ignore the presence of BCE-registered synthetic beans during unused beans removal, because any BCE-registered synthetic bean would automatically disable unused beans removal, because `SyntheticBeanCreator` unconditionally gives users an `Instance<Object>`.

We could fix that bug, but I would argue using injection points here is a more elegant solution: we don't just validate that the IPs can be satisfied, we can also enforce that the `SyntheticBeanCreator` cannot obtain more than it declared. This is precisely what all other kinds of beans do as well, and if they need dynamic lookup, they need to inject `Instance<Object>` explicitly. I believe synthetic beans should be _more_ consistent with other beans, not less. In other words, PE's `produceWith()` is an exception, not a rule.

> > This is exactly what you're suggesting in
> > > I am not sure I am happy with preventing user from accessing other beans should they wish to
> 
> Still nope. You can already access any bean you want to via `Instance` all over the place. Adding a hard check here won't change that.

Providing an `Instance<Object>` to _all_ `SyntheticBeanCreator`s was a straightforward way to bridge BCEs with PEs, but other than that, it was a mistake. I would even argue it was a mistake when it got added to PEs via `produceWith()`, it's just that noone cared enough for precision.

> I guess I don't like that the two APIs (PE and BCE) are just diverging in their capabilities here - PE allows you to do the validation as well (you can add IPS which are used just for validation which is exactly like `withInjectionPoints`) but you can then resolve any bean while BCE is slapping your fingers eagerly.

PEs and BCEs have always been wildly different in their capabilities, notably because there's a rift between deployment time and runtime and only a limited form of objects can be transferred through that rift. This is not an issue -- it is what enables build-time implementations.

Also, stop focusing on validation; `withInjectionPoint()` is way more than that. It is exactly like declaring an IP in a managed bean or producer method. Validation is a nice side effect, but the core value is that it makes synthetic bean closer to other beans.

> Of course I understand that without this, the requirement to declare IPs isn't enforceable - I was just questioning the usability of the API from any implementation that does not optimize away unused beans. And assuming we didn't enforce the failure, impls that want to have this optimization could still attempt the resolution and if the bean isn't present, they could blow up with precise exception detailing on what went wrong.

If you like the previous API better, just say so :-)

impl/src/main/resources/tck-audit-cdi.xml
Ladicek: What you have here looks good to me. I wouldn't lose sleep over this :-)

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsAnnotationInfo/AnnotationInfoExtension.java
```diff
@@ -0,0 +1,42 @@
...
+public class AnnotationInfoExtension implements BuildCompatibleExtension {
+    private AnnotationInfo niceQualifier;
+
+    @Registration(types = NiceWidget.class)
+    public void registration(BeanInfo bean) {
+        niceQualifier = bean.qualifiers().stream()
+                .filter(q -> q.name().equals(Nice.class.getName()))
+                .findFirst()
+                .orElse(null);
+    }
```
Ladicek: Does this have any particular meaning? Because you can create an `AnnotationInfo` using `AnnotationBuilder` directly in `@Synthesis`.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsBackwardCompat/SyntheticInjectionsBackwardCompatTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.jboss.cdi.tck.AbstractTest;
+import org.jboss.cdi.tck.cdi.Sections;
+import org.jboss.cdi.tck.shrinkwrap.WebArchiveBuilder;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.jboss.test.audit.annotations.SpecAssertion;
+import org.jboss.test.audit.annotations.SpecVersion;
+import org.testng.annotations.Test;
+
+@SpecVersion(spec = "cdi", version = "5.0")
+public class SyntheticInjectionsBackwardCompatTest extends AbstractTest {
```
Ladicek: The existing tests for synthetic beans verify the legacy behavior, I think? You don't change these in this PR.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsDependent/SyntheticInjectionsDependentTest.java
```diff
@@ -0,0 +1,63 @@
...
+        assertEquals(TrackedBean.destroyedCount.get(), 0);
+
+        handle.destroy();
+
+        // After synthetic bean destruction:
+        // - TrackedBean created in creator should be destroyed
+        // - TrackedBean created in disposer is also destroyed (disposer dependents
+        //   are destroyed when disposal completes)
+        assertEquals(TrackedBean.createdCount.get(), 2);
+        assertEquals(TrackedBean.destroyedCount.get(), 2);
```
Ladicek, line 60: These 2 don't look right. The creator and disposer don't create/destroy the `TrackedBean` explicitly, so it behaves as a regular dependent instance of the synthetic bean. Therefore, it lives as long as the synthetic bean lives, and these values should both be `1`, not `2`. Unless I'm missing something...?

Ladicek:
Tests mostly LGTM, baring a few comments. One more general (not even asking to change, although I won't stop you :-) ): if the implementation class of a synthetic bean is `FooBar`, by convention, the creator is `FooBarCreator` and disposer is `FooBarDisposer`. A lot of tests in this PR don't follow that, which makes them harder to read.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsDependent/SyntheticInjectionsDependentTest.java
manovotn: Aha, so my expectation here differ. I would expect new instance of the dependent bean in both cases (and that's what I did in Weld impl too).
The reason being, it is aligned with what happens if you attempt the same in producer/disposer combination. There, you get a new dependent bean instance in both cases.

EDIT: And it would also be different to what you'd get in PE where a similar case will always lead to two different beans :thinking:

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsAnnotationInfo/AnnotationInfoExtension.java
manovotn: Done

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsBackwardCompat/SyntheticInjectionsBackwardCompatTest.java
manovotn: Fair point, removed this.

manovotn:
I have addressed the remarks excluding the one with dependent bean instance being shared between creator and disposer. While implementable, this would be different from all the other places in CDI which I think would be quite surprising.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsDependent/SyntheticInjectionsDependentTest.java
Ladicek: You are right and I was wrong. This is indeed how producers/disposers work and how synthetic beans should also work. What you have is correct here and the `SyntheticInjections` javadoc is incorrect.

I retract my comment here and will submit a spec PR to clarify.

Ladicek: Hold on... why do we still assert `assertEquals(TrackedBean.destroyedCount.get(), 0);` after the creation function completes? I'm utterly confused now. Let me re-read the spec once again.

Ladicek: OK, I got it. So for producer methods, all `@Dependent` objects injected into a producer method are dependent objects of the produced instance and so are destroyed when the instance is destroyed. For disposer methods, all `@Dependent` objects are destroyed when the invocation completes.

The same happens for synthetic beans, so this test is correct. Let me just once again take a look at the BCE javadoc to make sure this is aligned.

Azquelt: That looks right to me.

One difference between synthetic beans and producer methods that sticks out to me here, is that for the producer method I'd expect two separate injection points, one for the producer method and another for the disposer method, whereas in this synthetic bean case there's only one injection point but we obtain a bean for it twice. I think this difference is ok.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsInjectionPoint/SyntheticInjectionsInjectionPointTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    @SpecAssertion(section = Sections.SYNTHESIS_PHASE, id = "ci", note = "@Dependent synthetic bean can obtain InjectionPoint from SyntheticInjections")
+    public void testInjectionPointLookup() {
+        Instance<Object> lookup = getCurrentBeanContainer().createInstance();
+        Instance.Handle<PojoConsumer> handle = lookup.select(PojoConsumer.class).getHandle();
+        PojoConsumer consumer = handle.get();
+        assertNotNull(consumer.pojo);
+        assertNotNull(consumer.pojo.injectionPoint);
+        assertEquals(consumer.pojo.injectionPoint.getType(), SyntheticPojo.class);
```
Azquelt, line 48: We could test a few more things about this injection point.

Since this represents an the injection point in `PojoConsumer` that `SyntheticPojo` was injected into, all of the methods should work.

`getQualifiers` is likely to be useful if any of the qualifiers have `@Nonbinding` members, so we should that (either here or as part of the qualifiers test)

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsInjectionPoint/SyntheticInjectionsInjectionPointExtension.java
```diff
@@ -0,0 +1,30 @@
...
+public class SyntheticInjectionsInjectionPointExtension implements BuildCompatibleExtension {
+    @Synthesis
+    public void synthesize(SyntheticComponents syn) {
+        syn.addBean(SyntheticPojo.class)
+                .type(SyntheticPojo.class)
+                .scope(Dependent.class)
+                .withInjectionPoint(InjectionPoint.class, new Annotation[0])
+                .createWith(SyntheticPojoCreator.class)
+                .disposeWith(SyntheticPojoDisposer.class);
+    }
```
Azquelt, line 30: This tests injecting `InjectionPoint` into a synthetic bean.

We should also test injecting `InjectionPoint` into a bean which is injected into a synthetic bean. E.g.:

- `SyntheticPojo` declares an injection point of `MyService`
- `MyService` is a dependent-scoped managed bean which has  `@Inject InjectionPoint ip`
- `MyService` should be able to call `getBean`, `getType`, `getQualifiers`, `isDelegate` and `isTransient` on the `InjectionPoint` and get correct values
- `getAnnotated` and `getMember` should return `null`

Azquelt (approved):
I'm approving what's here, but I've noted some test cases I think it would be good to cover in the comments.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsInjectionPoint/SyntheticInjectionsInjectionPointTest.java
manovotn: Fair enough, I will change this test to make use of a qualifier using binding and non-binding value + assert various parts of the injected IP.

impl/src/main/java/org/jboss/cdi/tck/tests/build/compatible/extensions/syntheticBeanInjectionsInjectionPoint/SyntheticInjectionsInjectionPointExtension.java
manovotn: I'll add a test for this as well :+1:

Ladicek (approved):

manovotn:
I have added a separate commit (to be squashed) with the tests suggested by @Azquelt.
Let me know if that hits the spot.

Azquelt (approved):
Thanks :+1:

PR #709, opened 2026-04-17
 add TCK tests for auto-closeable beans (`@AutoClose`)

Ladicek:
Fixes #700


Ladicek:
This is _no longer_ a draft ~~for 2 reasons~~ ~~because~~:

- ~~the PR includes commits from #659 (because it's a lot easier to avoid conflicts)~~
- ~~there are no tests for the `Producer` part of Portable Extensions~~

Other than that, it should be fine. Submitting it now for @manovotn's ongoing work.

Ladicek:
Added tests for `Producer`, but note I didn't run them, because they only make sense for CDI Full.

Ladicek:
Added a test that the call to `AutoCloseable.close()` during bean destruction is not a business method invocation. This is ready now.

The commits in this PR could be squashed, but don't necessarily have to be; I don't mind.

Ladicek:
Rebased, this can be merged now as there's no `SNAPSHOT` dependency anymore. Of course, after it's reviewed :-)

manovotn (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/autoclose/bean/AutoCloseBeanTest.java
```diff
@@ -0,0 +1,67 @@
...
+    public void testNoAutoClose() {
+        Bean<WithoutAutoCloseBean> bean = getUniqueBean(WithoutAutoCloseBean.class);
+        CreationalContext<WithoutAutoCloseBean> cc = getCurrentBeanContainer().createCreationalContext(bean);
+        WithoutAutoCloseBean instance = bean.create(cc);
+        bean.destroy(instance, cc);
+        assertTrue(instance.destroyed);
+    }
+
+    @Test
+    @SpecAssertion(section = BEAN, id = "bg")
```
manovotn, line 62: Just noticed - this should probably be id `bf`?
Same goes for the other 3 occurences of this ID in the test.

Ladicek: In my local `tck-audit-cdi.xml`, the relevant section reads:

```xml
<assertion id="bf">
    <text>|isEager()| must return |true| if the bean is eagerly initialized, and |false| otherwise.</text>
</assertion>

<assertion id="bg">
    <text>|isAutoClose()| must return |true| if the bean is auto-closeable, and |false| otherwise.</text>
</assertion>
```

So `bg` seems right here?

manovotn: Hm, you're right.
I must have misread that, sorry for the noise.

Ladicek:
Added one more commit that tests that auto-closeable dependent objects are closed when the containing bean is destroyed.

Azquelt (approved):
Looks good :+1:

I don't think we ever actually assert that `isAutoClose()` returns `false` anywhere, that may be worth adding.

Ladicek:
Good point, we test both for `isAlternative()`, `isReserve()` and `isEager()`, but not for `isAutoClose()`. I'll add something.

Ladicek:
Added an extra commit with tests for `Bean.isAutoClose()` returning `false`.

Azquelt (approved):

manovotn (approved):

manovotn:
I think we can merge this now - @Ladicek, I assume you want to squash those commits first?

Ladicek:
I didn't merge this because of the ongoing discussion in https://github.com/jakartaee/cdi/issues/968, but I guess we can merge anyway, because the spec was already merged too. Let me squash the commits and merge.

PR #718, opened 2026-05-12
Add deprecated compatibility bridges for renamed assertion methods

manovotn:
As per my comment in https://github.com/jakartaee/cdi-tck/pull/676#issuecomment-4435414321, this is what would be temporarily needed as a bridge before there is EE 12 release of platform TCKs with our changes (PR for that is https://github.com/jakartaee/platform-tck/pull/2703).

I am fine not going this way, I just created this as I was working out the cause so I though I might as well share.


Ladicek (approved):

Ladicek:
Makes sense to me. I didn't realize we have tests in the Platform TCK, my bad.

Azquelt (approved):

PR #724, opened 2026-05-20
improve injection points TCK tests

Ladicek:
TCK counterpart of https://github.com/jakartaee/cdi/pull/978

### DIFF
--- a/impl/src/main/java/org/jboss/cdi/tck/tests/full/definition/bean/custom/CustomBeanImplementationTest.java
+++ b/impl/src/main/java/org/jboss/cdi/tck/tests/full/definition/bean/custom/CustomBeanImplementationTest.java
@@ -60,6 +60,9 @@ public class CustomBeanImplementationTest extends AbstractTest {
     @SpecAssertions({ @SpecAssertion(section = DECLARING_SELECTED_ALTERNATIVES_BEAN_ARCHIVE, id = "h"),
             @SpecAssertion(section = INTER_MODULE_INJECTION, id = "q") })
     public void testGetBeanClassCalled() {
+        // note that `Bean.getBeanClass()` may be called multiple times on multiple places
+        // and this test therefore cannot verify it was called when verifying inter-module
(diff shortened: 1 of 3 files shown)


impl/src/main/java/org/jboss/cdi/tck/tests/full/definition/bean/custom/CustomBeanImplementationTest.java
```diff
@@ -138,15 +141,4 @@ public void testCustomBeanIsPassivationCapableDependency() throws IOException, C
...
-    @Test
-    @SpecAssertion(section = INTER_MODULE_INJECTION, id = "r")
-    public void testInjectionPointGetMemberIsUsedToDetermineTheClassThatDeclaresAnInjectionPoint() {
-        // resolve the bean and invoke a method on it to make sure it was created/used
-        Foo foo = getContextualReference(Foo.class, new PassivableLiteral());
-        foo.getId();
-        Assert.assertEquals(CustomInjectionPoint.getMembersClasses().size(), 2);
-        Assert.assertTrue(CustomInjectionPoint.getMembersClasses().contains(Bar.class));
-        Assert.assertTrue(CustomInjectionPoint.getMembersClasses().contains(Integer.class));
-    }
```
Ladicek, line 151: This looks weird on the first sight: why are we removing test coverage? The issue is: this doesn't really cover the stated intent. This only verifies that after obtaining a contextual reference, `InjectionPoint.getMember()` was called for each injection point. This may happen multiple times for multiple reasons -- for example, during deployment validation. If this test emptied the set at the beginning, it would fail later on on all implementations, because no implementation actually calls `InjectionPoint.getMember()` to verify inter-module visibility. Hence, the removal.

Azquelt (approved):
This makes sense to me, thanks.

manovotn:
Marked as ready for review as the related API PR was merged.

manovotn (approved):

PR #727, opened 2026-06-01
TCK tests for respecified wildcard type arguments for `Event` and `Instance`

Ladicek:
TCK counterpart for https://github.com/jakartaee/cdi/pull/982


Ladicek:
Fortunately, nearly all tests exercised cases that are still forbidden, I had to modify just one test.

Ladicek:
@manovotn had a good point -- I should also add tests for the meaningful cases, verifying that they do work. I'll do that.

Ladicek:
Updated.

manovotn:
> Updated.

I am going to toss in my review after I test this against my Weld PR (which I need to tweak as well) - probably tomorrow.

Ladicek:
Updated once again. `Event` may have a wildcard with lower bound as a type argument, `Instance` may have a wildcard with upper bound as a type argument.

I also reverted the change for CDI-232, because that change made the specification of the built-in bean for `Instance` a lot worse (all beans must have a well defined set of bean types!).

Ladicek:
Added two more commits for this PR to correspond to https://github.com/jakartaee/cdi/pull/982

Azquelt (approved):

Ladicek:
BTW I didn't update this PR to correspond to the latest changes in the spec PR, so I expect to look at this again.

Azquelt:
Ah sorry, you're right, is it just the text in the audit file that needs updating?

Ladicek:
That text definitely needs updating. I have to reevaluate if any tests need updates as well.

Ladicek:
I brought this up to date with the specification. You'll note there's just a superficial test for the new rules for `Instance` and `Event` bean assignability -- this is because this is extensively tested through other means, so I just added a bare minimum. If you want me to expand on it, I could be persuaded to :-)

manovotn (approved):
LGTM plus I have done a pass with Weld's pending PR and its working just fine.

Ladicek:
BTW I've just improved the superficial test for the new rules for `Instance` and `Event` bean assignability so it's not as superficial as it used to be.

impl/src/main/java/org/jboss/cdi/tck/tests/event/implicit/ImplicitEventTest.java
```diff
@@ -55,6 +63,33 @@ public static WebArchive createTestArchive() {
...
+        assertTrue(rawTypes.contains(Object.class));
+        assertTrue(rawTypes.contains(Event.class));
+    }
+
+    @Test
+    @SpecAssertion(section = BUILTIN_INSTANCE, id = "b")
+    public void testQualifiersOfBuiltinEvent() {
+        Bean<Event<StudentRegisteredEvent>> bean = getUniqueBean(STUDENT_REGISTERED_EVENT_LITERAL);
+        assertEquals(bean.getQualifiers(), Set.of(Any.Literal.INSTANCE, Default.Literal.INSTANCE));
+    }
```
Azquelt, line 91: I'm not sure we want to test this. I agree it matches the spec, but I still think the spec is more prescriptive here than it needs to be.

As discussed before, the actual bean types and qualifiers of `Event` and `Instance` shouldn't matter, as long as it can be injected everywhere that's required.

Implementing it as a bean with an infinite set of bean types and qualifiers should be fine, as should having some special case logic in the typesafe resolution.

manovotn: I am not sure I understand you @Azquelt.
Are you saying we shouldn't test type/qualifiers at all for this case or that we should make it more permissive and only check that the set contains *at least* `@Any` and `@Default` (for this particular test case)?

Ladicek: I don't quite get the objection here. We've respecified the `Instance` and `Event` beans to have a well-known finite sets of bean types and qualifiers (and all implementations I know of do this already), so why shouldn't we test it in the TCK?

(I confess I actually wrote the test for bean types such that the existing implementations pass without a change even though they don't really follow the spec to the letter. Want me to change it? :-) )

Azquelt: I don't think we should specify or test the types and qualifiers of these beans directly _at all_. I think I've been pretty consistent about arguing for this. My main argument against specifying an infinite set of types and qualifiers has always been that there's more than one valid way to do it so we shouldn't require one or the other. [[1](https://github.com/jakartaee/cdi/pull/982#issuecomment-4753482995), [2](https://github.com/jakartaee/cdi/pull/982#discussion_r3512480227)]

What we care about for `Event` and `Instance` is that they can be injected and looked for any valid qualifier and bean combination, which we did [by adding additional rules to typesafe resolution](https://github.com/jakartaee/cdi/blob/main/spec/src/main/asciidoc/core/injectionandresolution.asciidoc#performing-typesafe-resolution).

This is really a problem with the spec, not the test. I can open a PR there instead.

Ladicek: My main issue with this is that per https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#concepts (the very first section of the spec!), each bean has a set of bean types and a set of qualifiers. For built-in beans, those are obviously defined by the specification.

Azquelt: > For built-in beans, those are obviously defined by the specification.

The spec can either define things or leave it up to the implementation and there are examples of both in the spec for built-in beans.

We define there must be a built-in bean for both `BeanContainer` and `BeanManager`, but we don't define the complete set of bean types for either (implementations can add additional types), and we don't define whether they have to be the same bean or two separate beans.

In contrast we do define that in EE, the `BeanManager` implementation must also implement `ELAwareBeanManager`.

----

Looking at these tests, I don't _think_ the type tests are likely to break anyone (unless they did go down the route of implementing infinite sets), but the qualifier test should definitely allow additional qualifier types and probably shouldn't check for `@Default`:
- given that qualifiers are ignored during typesafe resolution, there's really no reason for the bean to have `@Default`
- it would also be reasonable for an implementation to use a special qualifier as an indicator that qualifiers should be ignored during typesafe resolution and to apply that to `Instance` and `Event`, so extra qualifiers should be permitted

Ladicek: > We define there must be a built-in bean for both `BeanContainer` and `BeanManager`, but we don't define the complete set of bean types for either

We don't define a _complete_ set of bean types anywhere, not even for `Instance` and `Event`. But we do define the minimum requirement, even for `BeanContainer` and `BeanManager` (and we have a bug there -- the `BeanManager` bean has to include `BeanContainer` in its bean types).


> the qualifier test should definitely allow additional qualifier types

Yeah, I think I agree with that.

> and probably shouldn't check for `@Default`

I'm not sure about this. Either we get rid of that statement in the spec, or we test it. I probably don't mind either, but all of ArC, Weld and OWB do have `@Default` in these beans. (I couldn't find how ODI defines these beans, sorry.)

Ladicek: > > the qualifier test should definitely allow additional qualifier types
> 
> Yeah, I think I agree with that.

Done.

manovotn (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/event/implicit/ImplicitEventTest.java
Azquelt: > I'm not sure about this. Either we get rid of that statement in the spec, or we test it. I probably don't mind either, but all of ArC, Weld and OWB do have `@Default` in these beans. (I couldn't find how ODI defines these beans, sorry.)

Ok, fair

> Done.

Thanks

Azquelt (approved):

impl/src/main/java/org/jboss/cdi/tck/tests/event/implicit/ImplicitEventTest.java
Ladicek: FTR, I looked into ODI and it does behave in a somewhat unexpected (albeit compliant) manner. The `Event` and `Instance` beans internally only have the `@Any` qualifier, but ODI achieves compliance (of a different part of the spec) by implementing the `Bean.getQualifiers()` method such that the result always contains `@Any` and if the interim result only has `@Any` and `@Named`, the final result includes `@Default` as well. So the end behavior is the same: these built-in beans both have `@Any` and `@Default`.

PR #730, opened 2026-06-04
improve signature testing

Ladicek:
The statically generated signature file in `impl/src/main/resources` is gone. The signature file is now generated during each build, and instead of attaching the static file using the Build Helper Maven plugin, the generated file is attached automatically by the Sigtest Maven plugin.

Further, a few more static files were removed, because they are not useful. The only actually useful file was moved from `impl` to `dist-build`, because that's the only use site.

The AsciiDoc documents in `dist-build` were slightly edited for better formatting and to remove typos, but I didn't do any major surgery.


dist-build/sigtest-weld.pom
```diff
@@ -0,0 +1,78 @@
...
+        <artifactId>project</artifactId>
+        <version>2.0.2</version>
+    </parent>
+
+    <groupId>jakarta.cdi</groupId>
+    <artifactId>cdi-tck-sigtest</artifactId>
+    <version>5.0</version>
+
+    <properties>
+        <weld.version>7.0.0.Alpha1</weld.version>
```
manovotn: Beta1 is the latest ATM. Might as well use that.

dist-build/sigtest-weld.pom
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.eclipse.ee4j</groupId>
+        <artifactId>project</artifactId>
+        <version>2.0.2</version>
```
manovotn: 2.0.4 is the latest I think - I'd use that.

manovotn (commented):
Just two small notes, otherwise looks good.

dist-build/sigtest-weld.pom
Ladicek: Done.

dist-build/sigtest-weld.pom
Ladicek: Done.

doc/reference/src/main/asciidoc/sigtest.asciidoc
```diff
@@ -2,150 +2,124 @@
...
+                            <goal>copy</goal>
+                        </goals>
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>jakarta.cdi</groupId>
+                                    <artifactId>jakarta.cdi-tck-core-impl</artifactId>
+                                    <version>${version.cdi-tck}</version>
+                                    <type>sigfile</type>
+                                    <classifier>jdk17</type>
```
manovotn:
```suggestion
                                    <classifier>jdk17</classifier>
```

manovotn:
We had some back and forth with Ladislav over Zulip and I think this version is good to go.
At least we have ended up preparing an automated sigtest check for both, Arc and Weld, so that's something :)

Thanks for delving into this @Ladicek!

doc/reference/src/main/asciidoc/sigtest.asciidoc
Ladicek: OMG :facepalm: Will fix.

Ladicek: Done. Also squashed the commits.

manovotn (approved):
As far as I can tell, this is looking good 👍

PR #741, opened 2026-08-03
Update CDI API to 5.0.0 and fix CI staging profile

manovotn:
The `staged` profile is what's needed to work off of staged Jakarta artifacts.

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -22,9 +22,6 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
-      - name: "Maven install(staging)"
-        run: |
-          mvn clean install -Dno-format -Pstaging -B -V
       - name: "Maven install"
         run: |
-          mvn clean install -Dno-format -B -V javadoc:javadoc
+          mvn clean install -Dno-format -Pstaged -B -V javadoc:javadoc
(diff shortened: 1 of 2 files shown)


Ladicek (approved):

manovotn:
I've added `-Pstaged` into both builds so the CI should now pass. I see no harm in having the profile active all the time (it is inherited from parent project and just activates Jakarta staging repos).

.github/workflows/ci-actions.yml
```diff
@@ -24,7 +24,7 @@ jobs:
           java-version: ${{ matrix.java }}
       - name: "Maven install(staging)"
         run: |
-          mvn clean install -Dno-format -Pstaging -B -V
+          mvn clean install -Dno-format -Pstaged -B -V
       - name: "Maven install"
         run: |
-          mvn clean install -Dno-format -B -V javadoc:javadoc
+          mvn clean install -Dno-format -Pstaged -B -V javadoc:javadoc
```
Ladicek, line 27: There's no point in having two builds that are identical, except one does `javadoc:javadoc` on top?

Ladicek (approved):

.github/workflows/ci-actions.yml
manovotn: That's what I thought I pushed...apparently, I didn't :facepalm: 
Should be fixed now - single build with `javadoc:javadoc` in it.

